PR #16, opened 2016-05-11
Add GCEInstanceTypesService for GCE branch.

baizhang:
This pull request includes the following changes:

1. Use GoogleCredentials to replace SignedJwtAssertionCredentials;
2. Add GCEInstanceTypesService;
3. Fix some tests for test_instance_types_service.py: methods find and get defined in interfaces/services.py return an instance object or None if not found, whereas the test cases expected the return value to be a list

This pull request passed the test_instance_types_service test. Environment variables GCE_PROJECT_NAME and GCE_DEFAULT_ZONE need to be set for the tests.


baizhang:
Add @afgane @nuwang @chiniforooshan @mbookman

mbookman:
LGTM

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -42,3 +43,48 @@ def material(self):
...
+    def size_ephemeral_disks(self):
+        return int(self._inst_dict.get('maximumPersistentDisksSizeGb'))
+
+    @property
+    def num_ephemeral_disks(self):
+        return self._inst_dict.get('maximumPersistentDisks')
+
+    @property
+    def extra_data(self):
+        return {key: val for key, val in enumerate(self._inst_dict)
```
nuwang: @baizhang Currently, the enumerate(self._inst_dict) generates a <sequence, key> pair, instead of a <key,value> pair. I think changing this to _inst_dict.items() would work in a python 2/3 compatible manner.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -73,18 +74,10 @@ def gce_compute(self):
 
     def _connect_gce_compute(self):
         if self.credentials_file:
-            with open(self.credentials_file) as f:
-                data = json.load(f)
-                credentials = SignedJwtAssertionCredentials(
-                    data['client_email'], data['private_key'],
-                    'https://www.googleapis.com/auth/compute')
+            credentials = GoogleCredentials.from_json(self.credentials_file)
```
nuwang: I couldn't get this branch of the code to work. I trid providing the credentials file as a bytearray, json object, and a filepath - but it complained on each go. I managed to get the get_application_default() branch to work by exporting the GOOGLE_APPLICATION_CREDENTIALS env var.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -73,18 +74,10 @@ def gce_compute(self):
 
     def _connect_gce_compute(self):
         if self.credentials_file:
-            with open(self.credentials_file) as f:
-                data = json.load(f)
-                credentials = SignedJwtAssertionCredentials(
-                    data['client_email'], data['private_key'],
```
nuwang: @baizhang Looks like the self.private_key property is no longer needed and can be required from the constructor altogether?

test/test_instance_types_service.py
```diff
@@ -102,7 +102,7 @@ def test_instance_types_get(self):
             self.provider,
             "instance_type")
         inst_type = self.provider.compute.instance_types.find(
-            name=instance_type_name)[0]
```
nuwang: @baizhang Throughout CloudBridge, the find method returns an array, since multiple matches are possible during a find. The get method in contrast, was intended for retrieving a single instance. Unfortunately, the docstring for the InstanceTypesService.find() method is incorrect, sorry about that, and states that the find() method returns an instance, instead of a list. Therefore, I think this should be fixed in the docstring instead.

nuwang:
@baizhang I tested it out and the InstanceTypeService works great! I've added some comments about some issues I encountered against the relevant line numbers.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -11,8 +11,9 @@
 
 from googleapiclient import discovery
 import httplib2
```
nuwang: @baizhang This import seems to be no longer needed

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -1,14 +1,18 @@
 from cloudbridge.cloud.base.resources import ClientPagedResultList
+from cloudbridge.cloud.base.services import BaseComputeService
+from cloudbridge.cloud.base.services import BaseInstanceTypesService
 from cloudbridge.cloud.base.services import BaseKeyPairService
 from cloudbridge.cloud.base.services import BaseSecurityGroupService
 from cloudbridge.cloud.base.services import BaseSecurityService
+from cloudbridge.cloud.interfaces.resources import InstanceType
```
nuwang: @baizhang  Possibly unused import?

cloudbridge/cloud/providers/gce/resources.py
baizhang: Done. Thanks for the catch!

cloudbridge/cloud/providers/gce/provider.py
baizhang: Done. I now use ServiceAccountCredentials to handle json file for the service account key. This is intended to be used only for cases where GoogleCredentials (the preferred solution) is not applicable.

The service account key json file I used for testing was downloaded from:
console.cloud.google.com --> APIs --> Credentials --> Create credentials | Service account key --> Compute Engine default service account

cloudbridge/cloud/providers/gce/provider.py
baizhang: Done.

test/test_instance_types_service.py
baizhang: Done.
Reverted the changes in test_instance_types_service.py and fixed the docstring.

cloudbridge/cloud/providers/gce/services.py

baizhang:
@nuwang Thank you very much for your review. I have fixed these issues (see inline comments). Please take another look.

nuwang:
@baizhang Looks great! Thanks for these changes. The only suggestion I have left is that perhaps we should use ServiceAccountCredentials.from_json_keyfile_dict instead of from_json_keyfile_name (https://github.com/google/oauth2client/blob/master/oauth2client/service_account.py). The reason is that it would help web scenarios where the user might upload the credentials file (in which case we don't need to write it to disk first before passing it to cloudbridge), and also fit in nicely with CloudBridge's config dict.  I suppose we'd need to rename the GCE_SERVICE_CREDS_FILE property to GCE_SERVICE_CREDS_DICT to match.

PR #17, opened 2016-05-27
Add GCERegionService for GCE branch

baizhang:

This pull request includes the following changes:

1. Add GCERegionService and related classes;
2. Use ServiceAccountCredentials.from_json_keyfile_dict instead of from_json_keyfile_name, so that credentials_dict can be passed in directly from config (per the comment in Pull Request #16).

This pull request passed the test_region_service.py.

### DIFF
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -31,6 +31,14 @@ class GCECloudProvider(BaseCloudProvider):
             'gce_project_name', os.environ.get('GCE_PROJECT_NAME'))
         self.credentials_file = self._get_config_value(
             'gce_service_creds_file', os.environ.get('GCE_SERVICE_CREDS_FILE'))
+        self.credentials_dict = self._get_config_value(
+            'gce_service_creds_dict', {})
+        # If 'gce_service_creds_dict' is not passed in from config and
+        # self.credentials_file is available, read and parse the json file to
+        # self.credentials_dict.
+        if self.credentials_file and not self.credentials_dict:
+            with open(self.credentials_file) as creds_file:
+                self.credentials_dict = json.load(creds_file)
         self.default_zone = self._get_config_value(
             'gce_default_zone', os.environ.get('GCE_DEFAULT_ZONE'))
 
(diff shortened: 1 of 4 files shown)


nuwang:
@baizhang Works great! 👍  I was thinking about the credential handling a bit more and realised that there are some interesting options that the new from_json_keyfile_dict method enables - so it would be great to know what you think. At present the GCE_PROJECT_NAME property is required. However, when the keyfile_dict is passed in, the dict internally contains the PROJECT_NAME. We could possibly extract the value automatically and thus make the GCE_PROJECT_NAME optional. However, I'm guessing that the default_credentials method will always require the GCE_PROJECT_NAME to be explicitly specified. So what's the best thing to do here -  is one design decision we will need to make.

The other is to decide whether to continue supporting the GCE_SERVICE_CREDS_FILE property. My thinking is that perhaps we could have a prioritised order as follows.
1. GCE_SERVICE_CREDS_FILE
2. GCE_SERVICE_CREDS_DICT
3. GCE default credentials.
What do you think?

baizhang:
Hi @nuwang 

Thanks for your review and merging the pull request!

The current priority order of credential handling is
1. If GCE_SERVICE_CREDS_DICT is provided in config, use GCE_SERVICE_CREDS_DICT;
2. If GCE_SERVICE_CREDS_DICT is not provided, but GCE_SERVICE_CREDS_FILE is set, parse  GCE_SERVICE_CREDS_FILE into credentials_dict;
3. If none of the above, use GCE default credentials.

My understanding is that using GCE default credentials is the recommended approach. In this pull request, I did not remove the option of using GCE_SERVICE_CREDS_FILE, because I found it quite convenient to load the credentials file specified by environment variable (at least for testing and development purposes).

@mbookman is the expert on the best practice of GCE credentials. Matt, could you please comment on the design of credential handling?

Thanks,
Bai

mbookman:
Providing multiple options (if all considered useful) is fine for the credentials passing, however the preferred approach that we have discussed is to:

- Only use the end user credentials for the creation of the Galaxy master node on Google Compute Engine.
- When a user connects to the Galaxy master node to change the cluster size nodes and their associated disk resources would be created using the credentials of the Compute Engine service account.

This avoids passing the end user credentials to the Galaxy master as well as saving them on the Galaxy master.

PR #19, opened 2016-06-27
GCE Security Groups Based on Firewalls

chiniforooshan:
This is an implementation of security groups based on GCE's network firewalls. Since GCE doesn't have the concept of a security group, they are simulated by GCE tags.

GCESecurityGroupService, GCESecurityGroup, and GCESecurityGroupRule keep minimal local data so that they don't get out of sync with server status; basically, they just keep the relevant tag name or firewall ID and GCESecurityGroupDelegate does all the hard work.

@baizhang @nuwang @afgane @mbookman

### DIFF
--- a/cloudbridge/cloud/providers/gce/README.rst
+++ b/cloudbridge/cloud/providers/gce/README.rst
@@ -1,0 +1,30 @@
+CloudBridge support for Google compute engine and cloud storage.
+
+Security Groups
+~~~~~~~~~~~~~~~
+CloudBridge API lets you control incoming traffic to VM instances by creating
+security groups, adding rules to security groups, and then assigning instances
+to security groups.
+
+GCE does this a little bit differently. GCE lets you assign `tags`_ to VM
+instances. Tags, then, can be used for networking purposes. In particular, you
+can create `firewall rules`_ to control incoming traffic to instances having a
+specific tag. So, to add GCE support to CloudBridge, we simulate security groups
+by tags.
+
+To make this more clear, let us consider the example of adding a rule to a
+security group. When you add a security group rule from the CloudBridge API to a
+security group ``sg``, what really happens, when the cloud provider is GCE, is
+that a firewall with one rule is created whose ``targetTags`` is ``[sg]``. This
+makes sure that the rule applies to all instances that have ``sg`` as a tag (in
+CloudBridge language instances belonging to the security group ``sg``).
+
+**Note**: This implementation does not take advantage of the full power of GCE
+firewall format and only creates firewalls with one rule and only sees firewalls
+with one rule.  This should be OK as long as all firewalls are created through
+the CloudBridge API.
+
+.. _`tags`: https://cloud.google.com/compute/docs/reference/latest/instances/
+   setTags
+.. _`firewall rules`: https://cloud.google.com/compute/docs/
+   networking#firewall_rules
--- a/cloudbridge/cloud/providers/gce/README.rst
+++ b/cloudbridge/cloud/providers/gce/README.rst
@@ -1,4 +1,6 @@
-CloudBridge support for Google compute engine and cloud storage.
+CloudBridge support for `Google Cloud Platform`_. Compute is provided by `Google
+Compute Engine`_ (GCE). Object storage is provided by `Google Cloud Storage`_
+(GCE).
 
 Security Groups
 ~~~~~~~~~~~~~~~
@@ -20,10 +22,15 @@ makes sure that the rule applies to all instances that have ``sg`` as a tag (in
 CloudBridge language instances belonging to the security group ``sg``).
 
 **Note**: This implementation does not take advantage of the full power of GCE
-firewall format and only creates firewalls with one rule and only sees firewalls
-with one rule.  This should be OK as long as all firewalls are created through
-the CloudBridge API.
+firewall format and only creates firewalls with one rule and only can find or
+list firewalls with one rule. This should be OK as long as all firewalls are
+created through the CloudBridge API.
 
+**Note**: The current implementation adds firewalls to the ``default`` network.
+
+.. _`Google Cloud Platform`: https://cloud.google.com/
+.. _`Google Compute Engine`: https://cloud.google.com/compute/docs
+.. _`Google Cloud Storage`: https://cloud.google.com/storage/docs
 .. _`tags`: https://cloud.google.com/compute/docs/reference/latest/instances/
    setTags
 .. _`firewall rules`: https://cloud.google.com/compute/docs/
--- a/cloudbridge/cloud/providers/gce/README.rst
+++ b/cloudbridge/cloud/providers/gce/README.rst
@@ -15,11 +15,11 @@ specific tag. So, to add GCE support to CloudBridge, we simulate security groups
 by tags.
 
 To make this more clear, let us consider the example of adding a rule to a
-security group. When you add a security group rule from the CloudBridge API to a
-security group ``sg``, what really happens, when the cloud provider is GCE, is
-that a firewall with one rule is created whose ``targetTags`` is ``[sg]``. This
-makes sure that the rule applies to all instances that have ``sg`` as a tag (in
-CloudBridge language instances belonging to the security group ``sg``).
+security group. When you add a security group rule from the CloudBridge API to
+a security group ``sg``, what really happens is that a firewall with one rule
+is created whose ``targetTags`` is ``[sg]``. This makes sure that the rule
+applies to all instances that have ``sg`` as a tag (in CloudBridge language
+instances belonging to the security group ``sg``).
 
 **Note**: This implementation does not take advantage of the full power of GCE
 firewall format and only creates firewalls with one rule and only can find or
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -7,7 +7,12 @@ from cloudbridge.cloud.base.resources import BasePlacementZone
 from cloudbridge.cloud.base.resources import BaseRegion
 from cloudbridge.cloud.base.resources import BaseSecurityGroup
 from cloudbridge.cloud.base.resources import BaseSecurityGroupRule
-from sets import Set
+
+# Older versions of Python do not have a built-in set data-structure.
+try:
+    set
+except NameError:
+    from sets import Set as set
 
 import hashlib
 import inspect
@@ -188,7 +193,7 @@ class GCEFirewallsDelegate(object):
 
     @property
     def tags(self):
-        out = Set()
+        out = set()
         for firewall in self.iter_firewalls():
             out.add(firewall['targetTags'][0])
         return out
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -220,6 +220,10 @@ class GCEFirewallsDelegate(object):
         if self.find_firewall(tag, ip_protocol, port, source_range,
                               source_tag) is not None:
             return True
+        # Do not let the user accidentally open traffic from the world by not
+        # explicitly specifying the source.
+        if source_tag is None and source_range is None:
+            return False
         firewall_number = 1
         suffixes = []
         for firewall in self.iter_firewalls(tag):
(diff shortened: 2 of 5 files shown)


cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -162,3 +168,327 @@ def zones(self):
...
                 for zone in zones]
+
+
+class GCEFirewallsDelegate(object):
+
+    _instance = None
+   
+    def __init__(self, provider):
+        if GCEFirewallsDelegate._instance is not None:
+            raise Exception('Use GCEFirewalls.instance() to instanciate.')
```
mbookman: s/instanciate/instantiate

mbookman:
Can you add a README such that the code reviewers (and end users) would understand  how Security Groups are being implemented here? One would want to know what Cloud objects should be created when deploying.

Thanks!

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -162,3 +168,327 @@ def zones(self):
...
+        self._provider = provider
+        self._list_response = None
+        GCEFirewallsDelegate._instance = self
```
nuwang: It looks like this class is meant to be a singleton? Will there be issues if there are multiple provider instances, each having different connection settings, and they call GCEFirewallsDelegate.get_instance() alternately, because it's caching the provider within the singleton?

nuwang:
@chiniforooshan Thanks for this pull request. Enis and I are both at a Galaxy conference, so it may take us a few days to review and merge this. Some quick comments inline in the meantime.

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: It's not a singleton anymore. For some reason I was under the impression that there is going to be only one provider, ever.

chiniforooshan:
@mbookman Do you think the added README file is clear enough?
@nuwang Thanks for looking at the pull request!

cloudbridge/cloud/providers/gce/README.rst
```diff
@@ -0,0 +1,30 @@
+CloudBridge support for Google compute engine and cloud storage.
```
mbookman:
```
CloudBridge support for `Google Cloud Platform`.
Compute is provided by `Google Compute Engine`_ (GCE).
Object storage is provided by `Google Cloud Storage`_ (GCS).

.. _Google Cloud Platform: https://cloud.google.com/
.. _Google Compute Engine: https://cloud.google.com/compute/docs
.. _Google Cloud Storage: https://cloud.google.com/storage/docs
```

cloudbridge/cloud/providers/gce/README.rst
```diff
@@ -0,0 +1,30 @@
...
+can create `firewall rules`_ to control incoming traffic to instances having a
+specific tag. So, to add GCE support to CloudBridge, we simulate security groups
+by tags.
+
+To make this more clear, let us consider the example of adding a rule to a
+security group. When you add a security group rule from the CloudBridge API to a
+security group ``sg``, what really happens, when the cloud provider is GCE, is
+that a firewall with one rule is created whose ``targetTags`` is ``[sg]``. This
+makes sure that the rule applies to all instances that have ``sg`` as a tag (in
+CloudBridge language instances belonging to the security group ``sg``).
```
mbookman: s/belonging/belong

cloudbridge/cloud/providers/gce/README.rst
```diff
@@ -0,0 +1,30 @@
...
+
+**Note**: This implementation does not take advantage of the full power of GCE
+firewall format and only creates firewalls with one rule and only sees firewalls
```
mbookman: I don't quite understand what "sees" means here.

cloudbridge/cloud/providers/gce/README.rst
```diff
@@ -0,0 +1,30 @@
...
+with one rule.  This should be OK as long as all firewalls are created through
+the CloudBridge API.
+
```
mbookman: It appears to me that in its current implementation, the only GCE network that this applies to is the "default" network. I think that at the very least, we should have the option of applying this to a user-specified GCE network.

This doesn't need to be blocking for the current pull request, but is something we should do in follow-up.

At the very least, the README should note that the firewall rules are added to the `default` network.

cloudbridge/cloud/providers/gce/README.rst
afgane: I feel _belonging_ is the intended word here: i.e., live instances that have SG `sg` associated with them?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -5,7 +5,13 @@
 from cloudbridge.cloud.base.resources import BaseKeyPair
 from cloudbridge.cloud.base.resources import BasePlacementZone
 from cloudbridge.cloud.base.resources import BaseRegion
+from cloudbridge.cloud.base.resources import BaseSecurityGroup
+from cloudbridge.cloud.base.resources import BaseSecurityGroupRule
+from sets import Set
```
afgane: This is not going to work with Python 3.

afgane:
Other than the few left over inline comments, this is working pretty good!

For example, with the following rules defined in the web console
screen shot 2016-07-13 at 2 30 18 pm

CloudBridge extracts apparent security groups:
```
>>> for sg in gce.security.security_groups.list(): print(sg.name, sg.rules)
(u'cloudman', [<CBSecurityGroupRule: IP: tcp; from: 20; to: 21; grp: None>, <CBSecurityGroupRule: IP: tcp; from: 22; to: 22; grp: None>, <CBSecurityGroupRule: IP: tcp; from: 443; to: 443; grp: None>])
(u'http-server', [<CBSecurityGroupRule: IP: tcp; from: 80; to: 80; grp: None>, <CBSecurityGroupRule: IP: tcp; from: 80; to: 80; grp: None>])
```

Nice job.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -162,3 +168,321 @@ def zones(self):
...
+    def description(self):
+        if self._description is not None:
+            return self._description
+        for firewall in self._delegate.iter_firewalls(self._security_group):
+            if 'description' in firewall:
+                return firewall['description']
+        return None
+
+    @property
+    def rules(self):
```
afgane: This method does not seem to refresh the underlying object so if a change has been done from the web console for example, it won't be reflected in the local CloudBridge object until a rule is added or removed from the security group object.

mbookman:
Looking at the screen shot from @afgane's test:

https://cloud.githubusercontent.com/assets/338642/16815048/621fad62-4906-11e6-923d-7aecf5c5f652.png

The added firewall rules appear to open the respective ports to the internet, rather than explicitly keeping them within the GCE project's network.

I'm looking at the Source tag / IP range / Subnetworks column and I see it indicating `0.0.0.0/0`.
I think that we should be either setting that to the IP range of the network (see the `default-allow-internal` firewall rule), or we should be setting the source tag to the target tag value (`cloudman` in this case).

The former would allow any instance within the same GCE network to connect to the galaxy instances, while the latter would restrict the connections to the Galaxy instances.

afgane:
The IP range set to 0.0.0.0/0 was a result how I created the firewall rule - I did so explicitly (e.g., `sg.add_rule(ip_protocol='tcp', from_port=100, to_port=100, cidr_ip='0.0.0.0/0')`). 

Alternatively, a rule can be added to a 'group': `sg.add_rule('tcp', 50, 50, src_group=sg)`

And then we get the following: 
screen shot 2016-07-13 at 4 03 43 pm
 
The launcher will create appropriate groups/rule for Galaxy as part of the launch process. 

Or were you referring to something else?

mbookman:
@afgane - Great. So long as documentation about usage is clear. I don't want it to be easy for users to open up ports to the internet when they didn't mean to.

cloudbridge/cloud/providers/gce/README.rst
chiniforooshan: Yes, I was trying to say what Enis explained. Given that, is belonging the write word here or should I change it? (Not a native English speaker :p)

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Yeah, I was not sure about this... I had to make a decision about the level of caching we want here and I thought maybe refreshing the local data whenever something is changed by the API is reasonable. Of course we can refresh for all reads and writes in expense of slowing down the operations a little bit.

Note that, AFAIK, this implementation is similar to [the AWS implementation](https://github.com/boto/boto/blob/develop/boto/ec2/securitygroup.py): the "rules" property returns the local copy of the list of rules and the local copy is updated only when "authorize" or "revoke" is called.

So, what is your opinion? Should we refresh at all reads and writes?

cloudbridge/cloud/providers/gce/README.rst
chiniforooshan: Nice catch! I added a note to the README file mentioning that this implementation only adds firewalls to the default network. I'll make this work for custom networks in a separate patch, when I figure out how to associate firewalls to networks, which is not immediately clear to me from the API documentation.

cloudbridge/cloud/providers/gce/README.rst
chiniforooshan: Done.

cloudbridge/cloud/providers/gce/README.rst
chiniforooshan: Changed "... sees firewalls with one rule" to "... can find or list firewalls with one rule". Does it make sense now?

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Thanks! Fixed.

chiniforooshan:
Thanks all for reviewing!

@mbookman Added a check when creating a new firewall to prevent users from accidentally opening traffic from the world by not explicitly specifying the source.

cloudbridge/cloud/providers/gce/resources.py
afgane: Yeah, it's a tossup... _boto_ has an `update` method for most of the objects so it's possible to explicitly refresh an object. We don't. @nuwang I remember we had a discussion on this update topic and you were of the opinion we didn't need an `update` method. Do we just require fetching the object again to refresh it or should we implement an implicit refresh within each function? Our AWS implementation for example also does not refresh the object but returns a stale one of a change has been done outside CloudBridge.

cloudbridge/cloud/providers/gce/README.rst
```diff
@@ -0,0 +1,37 @@
...
+GCE does this a little bit differently. GCE lets you assign `tags`_ to VM
+instances. Tags, then, can be used for networking purposes. In particular, you
+can create `firewall rules`_ to control incoming traffic to instances having a
+specific tag. So, to add GCE support to CloudBridge, we simulate security groups
+by tags.
+
+To make this more clear, let us consider the example of adding a rule to a
+security group. When you add a security group rule from the CloudBridge API to a
+security group ``sg``, what really happens, when the cloud provider is GCE, is
```
mbookman: Can remove the ", when the cloud provider is GCE,". I think it is clear from context that this is a GCE-specific discussion.

chiniforooshan: Done.

mbookman:
LGTM

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: @afgane, @nuwang ping :)

Just wanted to make sure that you are not waiting on anything from me. I was thinking about fixing the issue that @mbookman mentioned, about not being able to add firewalls to non-default networks, in a separate PR.

afgane: Thanks @chiniforooshan - I was waiting for @nuwang to chime in on the potential use the `update` method but I guess we can work on that piece in a separate thread if we all decide a path like that is desirable. 
I'll merge this in now and we can work on other pieces in the coming days/weeks.
Thanks.

PR #20, opened 2016-07-22
Add GCEImageService.

baizhang:

This pull request adds GCEImageService and related classes.

Currently, the change does not pass test_image_service.py, since the test needs to bring up a test instance and GCEInstanceService has yet to be implemented.

I tested the GCEImageService and related attributes manually as follows:

`export GCE_PROJECT_NAME=ubuntu-os-cloud   # Using a public GCE project`

`from cloudbridge.cloud.factory import CloudProviderFactory, ProviderList`
`provider = CloudProviderFactory().create_provider(ProviderList.GCE, {})`
`# List all Ubuntu images`
`print(provider.compute.images.list())`
`# Find Ubuntu images by name`
`print(provider.compute.images.find('ubuntu-1604-xenial-v20160627'))`
`# Find no Ubuntu images matching 'ubuntu-1604-xenial-v20160628'`
`print(provider.compute.images.find('ubuntu-1604-xenial-v20160628'))`
`# Get Ubuntu image by name`
`image = provider.compute.images.get('ubuntu-1604-xenial-v20160627')`
`# List attributes of the image`
`image.id`
`image.name`
`image.description`
`image.state`
`image.refresh()`
`image.name`

I also tested on deleting a private image in a private GCE project.

### DIFF
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -3,8 +3,12 @@ DataTypes used by this provider
 """
 from cloudbridge.cloud.base.resources import BaseInstanceType
 from cloudbridge.cloud.base.resources import BaseKeyPair
+from cloudbridge.cloud.base.resources import BaseMachineImage
 from cloudbridge.cloud.base.resources import BasePlacementZone
 from cloudbridge.cloud.base.resources import BaseRegion
+from cloudbridge.cloud.interfaces.resources import MachineImageState
+
+import googleapiclient
 
 
 class GCEKeyPair(BaseKeyPair):
@@ -162,3 +166,79 @@ class GCERegion(BaseRegion):
                  if zone['region'] == self._gce_region['selfLink']]
         return [GCEPlacementZone(self._provider, zone['name'], self.name)
                 for zone in zones]
+
+
+class GCEMachineImage(BaseMachineImage):
+
+    IMAGE_STATE_MAP = {
+        'PENDING': MachineImageState.PENDING,
+        'READY': MachineImageState.AVAILABLE,
+        'FAILED': MachineImageState.ERROR
+    }
+
+    def __init__(self, provider, image):
+        super(GCEMachineImage, self).__init__(provider)
+        if isinstance(image, GCEMachineImage):
+            # pylint:disable=protected-access
+            self._gce_image = image._gce_image
+        else:
+            self._gce_image = image
+
+    @property
+    def id(self):
+        """
+        Get the image identifier.
+        :rtype: ``str``
+        :return: ID for this instance as returned by the cloud middleware.
+        """
+        return self._gce_image['id']
+
+    @property
+    def name(self):
+        """
+        Get the image name.
+        :rtype: ``str``
+        :return: Name for this image as returned by the cloud middleware.
+        """
+        return self._gce_image['name']
+
+    @property
+    def description(self):
+        """
+        Get the image description.
+        :rtype: ``str``
+        :return: Description for this image as returned by the cloud middleware
+        """
+        return self._gce_image['description']
+
+    def delete(self):
+        """
+        Delete this image
+        """
+        request = self._provider.gce_compute.images().delete(
+            project=self._provider.project_name, image=self.name)
+        request.execute()
+
+    @property
+    def state(self):
+        return GCEMachineImage.IMAGE_STATE_MAP.get(
+            self._gce_image['status'], MachineImageState.UNKNOWN)
+
+    def refresh(self):
+        """
+        Refreshes the state of this instance by re-querying the cloud provider
+        for its latest state.
+        """
+        try:
+            response = self._provider.gce_compute \
+                                  .images() \
+                                  .get(project=self._provider.project_name,
+                                       image=self.name) \
+                                  .execute()
+            if response:
+                # pylint:disable=protected-access
+                self._gce_image = response
+        except googleapiclient.errors.HttpError as http_error:
+            # image no longer exists
+            print("googleapiclient.errors.HttpError: {0}".format(http_error))
+            self._gce_image['status'] = "unknown"
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -7,6 +7,7 @@ from cloudbridge.cloud.base.resources import BaseMachineImage
 from cloudbridge.cloud.base.resources import BasePlacementZone
 from cloudbridge.cloud.base.resources import BaseRegion
 from cloudbridge.cloud.interfaces.resources import MachineImageState
+import cloudbridge as cb
 
 import googleapiclient
 import re
@@ -235,7 +236,7 @@ class GCEMachineImage(BaseMachineImage):
         if match:
             project = match.group(1)
         else:
-            print("Project name is not found.")
+            cb.log.warning("Project name is not found.")
             return
         try:
             response = self._provider.gce_compute \
@@ -248,5 +249,6 @@ class GCEMachineImage(BaseMachineImage):
                 self._gce_image = response
         except googleapiclient.errors.HttpError as http_error:
             # image no longer exists
-            print("googleapiclient.errors.HttpError: {0}".format(http_error))
+            cb.log.warning(
+                "googleapiclient.errors.HttpError: {0}".format(http_error))
             self._gce_image['status'] = "unknown"
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -9,7 +9,7 @@ from cloudbridge.cloud.base.resources import BaseRegion
 from cloudbridge.cloud.interfaces.resources import MachineImageState
 
 import googleapiclient
-
+import re
 
 class GCEKeyPair(BaseKeyPair):
 
@@ -229,10 +229,18 @@ class GCEMachineImage(BaseMachineImage):
         Refreshes the state of this instance by re-querying the cloud provider
         for its latest state.
         """
+        resource_link = self._gce_image['selfLink']
+        project_pattern = 'projects/(.*?)/'
+        match = re.search(project_pattern, resource_link)
+        if match:
+            project = match.group(1)
+        else:
+            print("Project name is not found.")
+            return
         try:
             response = self._provider.gce_compute \
                                   .images() \
-                                  .get(project=self._provider.project_name,
+                                  .get(project=project,
                                        image=self.name) \
                                   .execute()
             if response:
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -210,7 +210,7 @@ class GCEMachineImage(BaseMachineImage):
         :rtype: ``str``
         :return: Description for this image as returned by the cloud middleware
         """
-        return self._gce_image['description']
+        return self._gce_image.get('description', '')
 
     def delete(self):
         """
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -192,7 +192,7 @@ class GCEMachineImage(BaseMachineImage):
         :rtype: ``str``
         :return: ID for this instance as returned by the cloud middleware.
         """
-        return self._gce_image['id']
+        return self._gce_image['name']
 
     @property
     def name(self):
--- a/cloudbridge/cloud/providers/gce/services.py
+++ b/cloudbridge/cloud/providers/gce/services.py
@@ -1,5 +1,6 @@
 from cloudbridge.cloud.base.resources import ClientPagedResultList
 from cloudbridge.cloud.base.services import BaseComputeService
+from cloudbridge.cloud.base.services import BaseImageService
 from cloudbridge.cloud.base.services import BaseInstanceTypesService
 from cloudbridge.cloud.base.services import BaseKeyPairService
 from cloudbridge.cloud.base.services import BaseRegionService
@@ -12,6 +13,7 @@ import googleapiclient
 
 from retrying import retry
 
+from .resources import GCEMachineImage
 from .resources import GCEInstanceType
 from .resources import GCEKeyPair
 from .resources import GCERegion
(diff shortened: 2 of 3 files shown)


cloudbridge/cloud/providers/gce/services.py
```diff
@@ -267,16 +269,84 @@ def current(self):
...
+            return []
+        if 'items' not in response:
+            return []
+        images = [GCEMachineImage(self.provider, image) for image
+                  in response['items']
+                  if 'name' in image and image['name'] == filters['name']]
+        return ClientPagedResultList(self.provider, images,
+                                     limit=limit, marker=marker)
+
+    def list(self, limit=None, marker=None):
```
afgane: Since the GCE API appears to supports paging directly in the API (via `maxResults` param) so we should probably be using that instead of the `ClientPagedResultList`, which is just a library-internal workaround for the APIs that don't natively support paging.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -267,16 +269,84 @@ def current(self):
...
+        """
+        List all images.
+        """
+        try:
+            response = self.provider.gce_compute \
+                                  .images() \
+                                  .list(project=self.provider.project_name) \
```
afgane: We have another GCE-discrepancy from AWS|OS here. The `project` argument filters images for the given project. So, as currently implemented, I just get an empty list since it only inspects the current project. If I set the value of `project` arg to `debian-cloud` for example, I get a list of images from that project. I would think that the `list` image (as well as `find` and `get`) should look at all public and private images by default. That's the behavior for AWS & OS. How do we achieve this for GCE? From the web console, I see a list of 18 images covering all the basic operating systems. How is that retrieved? Alternatively, how can one get a list of projects on GCE? I found `debian-cloud` in the docs, but which other ones exist? Should we compose a list of projects whose images we list/use?

mbookman: The API for listing images is per project:
https://cloud.google.com/compute/docs/reference/latest/images/list

The list of publicly available projects can be found here:
https://cloud.google.com/compute/docs/images

It looks like clients just need to hard-code these public projects. See libcloud, for exampe:

https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/gce.py#L1051

afgane: Thanks for the pointers Matt - it's good to see the public images are officially documented. Given that, I would think a similar solution to what's in Libcloud makes sense here as well.

cloudbridge/cloud/providers/gce/services.py
baizhang: Since now we append public images to the end of project-specific private images, ClientPagedResultList is used here to provide a paged view of all available images.

cloudbridge/cloud/providers/gce/services.py
baizhang: I hard-coded five public image projects, 'centos-cloud', 'coreos-cloud', 'debian-cloud', 'opensuse-cloud', 'ubuntu-os-cloud', which provide non-premium images. When any of the methods of GCEImageService is first called, the public images will be retrieved and cached in _public_images of the GCEImageService object.

baizhang:
@afgane, thank you very much for reviewing this pull request!

I have added public images to GCEImageService. Please take another look.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -273,11 +273,33 @@ class GCEImageService(BaseImageService):
...
+            if 'items' in response:
+                self._public_images.extend(
+                    [GCEMachineImage(self.provider, image) for image
+                     in response['items']])
 
     def get(self, image_name):
         """
         Returns an Image given its name
         """
+        if self._public_images == None:
```
mbookman: Nit: Make this an "is" comparison.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -290,48 +312,54 @@ def get(self, image_name):
...
+        for public_image in self._public_images:
+            if public_image.name == image_name:
+                return public_image
+        return None
 
     def find(self, name, limit=None, marker=None):
         """
         Searches for an image by a given list of attributes
         """
+        if self._public_images == None:
```
mbookman: What do you think about always calling self._retrieve_public_images() and have the check for "None" in that function. That keeps the logic in one place of "None" means we haven't tried to fetch, "a list, even if empty" means we have.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -290,48 +312,54 @@ def get(self, image_name):
...
+                          in response['items']]
         except googleapiclient.errors.HttpError as http_error:
             print("googleapiclient.errors.HttpError: {0}".format(http_error))
-            return []
-        if 'items' not in response:
-            return []
-        images = [GCEMachineImage(self.provider, image) for image
-                  in response['items']
-                  if 'name' in image and image['name'] == filters['name']]
+        images.extend(self._public_images)
```
mbookman: This always adds the public images to the response, even if a name filter is provided?
If that is the correct logic, please add a comment here explaining it because it is not obviously correct.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -290,48 +312,54 @@ def get(self, image_name):
...
+                images = [GCEMachineImage(self.provider, image) for image
+                          in response['items']]
         except googleapiclient.errors.HttpError as http_error:
             print("googleapiclient.errors.HttpError: {0}".format(http_error))
-            return []
-        if 'items' not in response:
-            return []
-        images = [GCEMachineImage(self.provider, image) for image
-                  in response['items']]
+        images.extend(self._public_images)
```
mbookman: Does order of results matter? In find(), public images are added before project-level images. In list() public images are added after project-level images.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -273,11 +273,33 @@ class GCEImageService(BaseImageService):
 
     def __init__(self, provider):
         super(GCEImageService, self).__init__(provider)
+        self._public_images = None
+
+    PUBLIC_IMAGE_PROJECTS = ['centos-cloud', 'coreos-cloud', 'debian-cloud',
```
mbookman: If the PUBLIC_IMAGE_PROJECTS list is intended to be class private, then prefix this with an underscore.

cloudbridge/cloud/providers/gce/services.py

cloudbridge/cloud/providers/gce/services.py
baizhang: The next statement does the actual filtering.

cloudbridge/cloud/providers/gce/services.py
baizhang: The order is always project-specific images first and then public images.

baizhang:
@mbookman, thank you for your review comments!

Please take another look. Thanks!

afgane:
Thanks for the update. This looks nice - I only have some reservations about the `print` statements. @nuwang what's your take on this topic? I guess until we add custom exceptions, print's may be ok but I wonder if (re)raising an exception might be better (that's what we've done for the other providers).

nuwang:
+1 Looks great. Can we use the standard python logging module? A high priority item for the next release is to get logging properly sorted out and handled consistently throughout the library.

baizhang:
Hi @afgane @nuwang , thanks for the review comments! I made the change to use logging instead of print statements to track API exceptions.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -162,3 +167,88 @@ def zones(self):
...
+        return self._gce_image['name']
+
+    @property
+    def description(self):
+        """
+        Get the image description.
+        :rtype: ``str``
+        :return: Description for this image as returned by the cloud middleware
+        """
+        return self._gce_image['description']
```
nuwang: @baizhang Just wondering, is the 'description' guaranteed to be there? Or should it be _gce_image.get('description') just in case?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -162,3 +167,88 @@ def zones(self):
...
+    def state(self):
+        return GCEMachineImage.IMAGE_STATE_MAP.get(
+            self._gce_image['status'], MachineImageState.UNKNOWN)
+
+    def refresh(self):
+        """
+        Refreshes the state of this instance by re-querying the cloud provider
+        for its latest state.
+        """
+        resource_link = self._gce_image['selfLink']
```
nuwang: @baizhang What do you think about calling the ImageService's get method here with self.id? (as in the AWS/OpenStack implementation).

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -267,16 +270,122 @@ def current(self):
...
+                                        .execute()
+            except googleapiclient.errors.HttpError as http_error:
+                cb.log.warning("googleapiclient.errors.HttpError: {0}".format(
+                    http_error))
+            if 'items' in response:
+                self._public_images.extend(
+                    [GCEMachineImage(self.provider, image) for image
+                     in response['items']])
+
+    def get(self, image_name):
```
nuwang: @baizhang The method signature is incorrect I think. This should be get(self, image_id) and should retrieve by id.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -267,16 +270,122 @@ def current(self):
...
+        """
+        Returns an Image given its name
+        """
+        self._retrieve_public_images()
```
nuwang: @baizhang What do you think about moving the public image retrieval logic into the except block? That way, we  only hit it it if the get() fails on the private images.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -267,16 +270,122 @@ def current(self):
...
+                    return public_image
+            cb.log.warning(
+                "googleapiclient.errors.HttpError: {0}".format(http_error))
+        return None
+
+    def find(self, name, limit=None, marker=None):
+        """
+        Searches for an image by a given list of attributes
+        """
+        self._retrieve_public_images()
```
nuwang: @baizhang This could be moved into the else block? This would prevent hitting _retrieve_public_images() it if it's not a public project.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -267,16 +270,122 @@ def current(self):
...
+        filters = {'name': name}
+        images = []
+        if (self.provider.project_name not in
+            GCEImageService._PUBLIC_IMAGE_PROJECTS):
+            try:
+                response = self.provider \
```
nuwang: @baizhang This could potentially be simplified by iterating through self (as in OpenStack implementation). The self iterator calls the list method internally.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -267,16 +270,122 @@ def current(self):
...
+        images.extend(self._public_images)
+        images = [image for image in images if image.name == filters['name']]
+        return ClientPagedResultList(self.provider, images,
+                                     limit=limit, marker=marker)
+
+    def list(self, limit=None, marker=None):
+        """
+        List all images.
+        """
+        self._retrieve_public_images()
```
nuwang: @baizhang Could this call to _retrieve_public_images() be moved to the else clause?

nuwang:
@baizhang @mbookman @chiniforooshan
Have gone through it and tested the code and it works great. I see the problem you mentioned with the tests - that is, the image service test relies on the instance service, but the instance service doesn't exist yet. If you have any suggestions on how that test could be improved, that'd be a great help.

I've also added a few more suggestions inline. Will wait for your views on it and then merge everything in.

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/resources.py
baizhang: The reason I cannot simply call get method here is that I need to figure out which project this image resource belongs to. In GCE image API, both the project and the image name need to be specified for the get method.

cloudbridge/cloud/providers/gce/services.py
baizhang: In GCE API, the value of `name` is something like u'ubuntu-1604-xenial-v20160907a', while the value of `id` is something like 2932474445039421428.

According to https://cloud.google.com/compute/docs/reference/latest/images, id is "[Output Only] The unique identifier for the resource. This identifier is defined by the server." And get method uses image `name` to specify the resource.

cloudbridge/cloud/providers/gce/services.py

cloudbridge/cloud/providers/gce/services.py
baizhang: I have removed this line and used self.list method instead.

cloudbridge/cloud/providers/gce/services.py
baizhang: Done. I made an explicit call to self.list to set the limit to sys.maxsize. My understanding is the results will be truncated to provider.config.default_result_limit if I use the self iterator call, so we may miss some results. Please correct me if I'm mistaken here.

cloudbridge/cloud/providers/gce/services.py
baizhang: This doesn't seem to make much difference here, since we always include public images in list result.

baizhang:
@nuwang Thank you very much for your review comments! Please take another look.

cloudbridge/cloud/providers/gce/resources.py
nuwang: I see what you mean - so if my understanding is correct - the list of images are either from the project's image list, or the public image list. By using selfLink, you can handle both cases in one go - is that correct?

cloudbridge/cloud/providers/gce/services.py
nuwang: I see, so the GCE APIs use 'name' as the external primary key, and 'id' as some internal pk value which is not intended for public consumption? In that case, it may make sense to make the GCEMachineImage.id function return self._gce_image['name'] instead of self._gce_image['id']. This is because in CloudBridge, the id refers to the primary key through which an object can be retrieved.  Subsequently, the get method's signature can also use the image_id?

cloudbridge/cloud/providers/gce/services.py
nuwang: Sorry, it looks like I've commented in the wrong place. What I meant was in the get() method. _retrieve_public_images() only needs to be called if searching through the project's image list fails to turn up the image. Therefore, it can be moved to the except clause, where public images are searched?

cloudbridge/cloud/providers/gce/services.py
nuwang: What I meant was something like:
images = [image for image in self if image.name == filters['name']]

The [image in self] iterator works because every CloudBridge resource inherits an iterator from its parent class. The parent implementation (BasePageableObjectMixin) internally calls the list() method to page in the objects, and if you iterate past the end of the list, it will automatically call list again to get the next set and so on, allowing you to iterate through all available objects in one go.

cloudbridge/cloud/providers/gce/resources.py
baizhang: Yes.

cloudbridge/cloud/providers/gce/services.py
baizhang: I see. Thank you for the comment. I've made the change accordingly.

cloudbridge/cloud/providers/gce/services.py
baizhang: Done (in previous commit).

baizhang:
Hi @nuwang, thanks! I've made the changes based on your comments.

PR #21, opened 2016-08-25
Gce custom net

chiniforooshan:
Hi everyone,

This PR merges recent changes on master into the GCE branch and also adds supports for creating security groups in custom networks.

@mbookman, @baizhang, @afgane, @nuwang

### DIFF
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -429,7 +429,7 @@ class GCESecurityGroup(BaseSecurityGroup):
         network and the target tag corresponding to this security group.
         """
         return GCEFirewallsDelegate.tag_network_id(self._security_group,
-                                                   self.network)
+                                                   self._network)
 
     @property
     def name(self):
@@ -451,7 +451,7 @@ class GCESecurityGroup(BaseSecurityGroup):
         if self._description is not None:
             return self._description
         for firewall in self._delegate.iter_firewalls(self._security_group,
-                                                      self.network):
+                                                      self._network):
             if 'description' in firewall:
                 return firewall['description']
         return None
@@ -460,13 +460,9 @@ class GCESecurityGroup(BaseSecurityGroup):
     def rules(self):
         out = []
         for firewall in self._delegate.iter_firewalls(self._security_group,
-                                                      self.network):
+                                                      self._network):
             out.append(GCESecurityGroupRule(self._delegate, firewall['id']))
         return out
-
-    @property
-    def network(self):
-        return self._network
 
     @staticmethod
     def to_port_range(from_port, to_port):
@@ -483,7 +479,7 @@ class GCESecurityGroup(BaseSecurityGroup):
         src_tag = src_group.name if src_group is not None else None
         self._delegate.add_firewall(self._security_group, ip_protocol, port,
                                     cidr_ip, src_tag, self.description,
-                                    self.network)
+                                    self._network)
         return self.get_rule(ip_protocol, from_port, to_port, cidr_ip,
                              src_group)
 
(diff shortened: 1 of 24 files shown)


cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -176,66 +177,80 @@ def zones(self):
...
+    NETWORK_URL_PREFIX = 'global/networks/'
+    DEFAULT_NETWORK = 'default'
   
     def __init__(self, provider):
         self._provider = provider
         self._list_response = None
 
     @staticmethod
-    def tag_id(tag):
+    def tagnet_id(tag, network):
```
mbookman: "tagnet" doesn't immediately map to a concept I recognize. Is there a better name for this (it isn't used much in here, so it could be more verbose)? If not, please add a comment as to what a "tagnet" represents.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -176,66 +177,80 @@ def zones(self):
 
 
 class GCEFirewallsDelegate(object):
+    NETWORK_URL_PREFIX = 'global/networks/'
```
mbookman: Are these two "constants" intended to be class private? If so, prefix them with an underscore.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Sure. Changed it to "tag_network" to be more clear and added inline documentation.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: NETWORK_URL_PREFIX is class private and so I prefixed it with an underscore. DEFAULT_NETWORK is used in the GCESecurityGroup class, too.

chiniforooshan:
@afgane, @nuwang Could you please take a look? Thanks!

test/test_security_service.py
```diff
@@ -103,7 +99,7 @@ def test_crud_security_group_service(self):
         sg = self.provider.security.security_groups.create(
             name=name, description=name)
         #Empty security groups don't exist in GCE. Let's add a dummy rule.
```
nuwang: @chiniforooshan @mbookman @afgane I think that we need to capture this constraint at the interface level and remove the line which manually adds a dummy rule in the test. Right now, the GCE behaviour is different from the AWS and OS implementations. Calling SecurityGroup.Create() results in a new resource being created in the latter, but a dummy wrapper being returned in the former. An actual security group will only be created if you add at least one rule in GCE. This means that a user who calls create() and expects the group to exist, will get unexpected behaviour in GCE.

One solution could be to change the create method's signature at the interface level to create(sg_name, initialRules, description=none), where initialRules is a non-null set of rules that can be specified when creating the securityGroup. This ensures that it's impossible, by design, to create a security group without at least a single rule.

We could also consider radically different designs. My understanding of GCE  rules though is that, rules (a.k.a firewalls) exist in isolation, and can be tagged. These tags can subsequently be assigned to multiple instances, in effect applying the rule. Ehsan's implementation groups these rules by tag to simulate a SecurityGroup. Is that correct? If you remove a rule, only the tag is removed? What happens when all tags are removed, is the rule automatically deleted by GCE?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -196,30 +196,34 @@ def __init__(self, provider):
...
+        tag, network = self._delegate.get_tag_network_from_id(group_id)
+        if tag is None:
+            return None
+        return GCESecurityGroup(self._delegate, tag, network)
 
     def list(self, limit=None, marker=None):
-        security_groups = [GCESecurityGroup(self._delegate, x)
-                           for x in self._delegate.tags]
+        security_groups = [GCESecurityGroup(self._delegate, x, y)
+                           for x, y in self._delegate.tag_networks]
```
nuwang: I'd like to suggest calling x-> name and y-> network_id for clarity.

test/test_security_service.py
afgane: Re. the change of the signature for SG's `create` method - adding the `initialRules` parameter sounds like a good idea; I don't see a realistic usecase for an empty SG anyhow.

PR #24, opened 2016-11-04
Network service

chiniforooshan:
This is part 1 of the GCE network service implementation. There are 3 more remaining parts.
    
The test for network service is broken into 4 smaller tests.

@baizhang @nuwang @afgane @mbookman

### DIFF
--- a/cloudbridge/cloud/providers/gce/services.py
+++ b/cloudbridge/cloud/providers/gce/services.py
@@ -3,6 +3,7 @@ from cloudbridge.cloud.base.services import BaseComputeService
 from cloudbridge.cloud.base.services import BaseImageService
 from cloudbridge.cloud.base.services import BaseInstanceTypesService
 from cloudbridge.cloud.base.services import BaseKeyPairService
+from cloudbridge.cloud.base.services import BaseNetworkService
 from cloudbridge.cloud.base.services import BaseRegionService
 from cloudbridge.cloud.base.services import BaseSecurityGroupService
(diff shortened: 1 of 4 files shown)


cloudbridge/cloud/providers/gce/services.py
```diff
@@ -417,3 +419,59 @@ def instances(self):
...
+        try:
+            response = (self.provider.gce_compute
+                                     .networks()
+                                     .insert(project=self.provider.project_name,
+                                             body={'name': name})
+                                     .execute())
+            if 'error' in response:
+                return None
+            self.provider.wait_for_global_operation(response)
+            networks = self.list(filter='name eq %s' % name)
```
nuwang: I think we need to come up with a way to standardise the filter vocabulary, since this won't work on non-GCE clouds. However, since the filter parameter is woefully underspecified in the cloudbridge spec, until we come up with a generalised solution, I propose we leave this as is for now.

test/test_network_service.py
```diff
@@ -60,45 +73,45 @@ def test_crud_network_service(self):
...
-                .format(ip_id))
-
         netl = self.provider.network.list()
         found_net = [n for n in netl if n.name == name]
         self.assertEqual(
             len(found_net), 0,
             "Network {0} should have been deleted but still exists."
             .format(name))
 
+    def test_crud_ip(self):
```
nuwang: @chiniforooshan Thanks for improving the tests. Looks great.

PR #26, opened 2017-01-23
Merge master into gce and resolve network id issues

chiniforooshan:
Instead of passing the network name, we now pass the network object
to the security group so that it has access to network name and
network id.

Also, there is a bug in the GCE API that in some cases, when the
network id has more than 19 digits, we cannot filter by id. As a
workaround, we list all networks and filter locally.

@baizhang @mbookman @nuwang @afgane

### DIFF
--- a/test/test_security_service.py
+++ b/test/test_security_service.py
@@ -1,3 +1,4 @@
+"""Test cloudbridge.security modules."""
 import json
 from test.helpers import ProviderTestBase
 import time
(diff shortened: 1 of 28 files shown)


test/test_security_service.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, methodName, provider):
             methodName=methodName, provider=provider)
 
     def test_crud_key_pair_service(self):
-        name = 'cbtestkeypair-a'
+        name = 'cbtestkeypairA-{0}'.format(uuid.uuid4()).lower()
```
nuwang: @chiniforooshan Is this to prevent collisions between test runs?

chiniforooshan: Thanks for reviewing. This renaming is not really needed since the test security groups and networks are deleted after running each test. I just wanted to make it more like what we have on the master branch.

PR #27, opened 2017-01-31
Add GCEInstanceService.

baizhang:

This pull request adds GCEInstanceService and related classes.

I tested the GCEInstanceService and related attributes manually as follows:

`from cloudbridge.cloud.factory import CloudProviderFactory, ProviderList`
`provider = CloudProviderFactory().create_provider(ProviderList.GCE, {})`
`image = provider.compute.images.find('ubuntu-1604-xenial-v20160627')[0]`
`instance_type = provider.compute.instance_types.list()[0]`

`# Set up security groups`
`net = provider.network.list()[0] # default network`
`sg = provider.security.security_groups.create('test-sg', 'test security',  net.id)`
`rule = sg.add_rule(src_group=sg, ip_protocol='tcp')`
`provider.security.security_groups.list()`

`provider.compute.instances.create('test-instance-123', image, instance_type, security_groups=['test-sg'])`
`provider.compute.instances.list()`
`instance_id = provider.compute.instances.find('test-instance-123', limit=10)[0].id`
`instance = provider.compute.instances.get(instance_id)`

`instance.id`
`instance.name`
`instance.security_groups`
`instance.instance_type_id`
`instance.instance_type`

`# Wait till the instance is running`
`instance.refresh()`
`instance.state`
`instance.public_ips`
`instance.private_ips`

`instance.stop()`
`# Wait a few seconds`
`instance.refresh()`
`instance.state`

`instance.reboot()`
`# Wait a few seconds`
`instance.refresh()`
`instance.state`

`instance.terminate()`
`# Wait a few seconds`
`provider.compute.instances.list()`

### DIFF
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -5,11 +5,13 @@ for GCE.
 
 
 from cloudbridge.cloud.base import BaseCloudProvider
+import httplib2
 import json
 import os
 import time
 
 from googleapiclient import discovery
+import googleapiclient.http
 from oauth2client.client import GoogleCredentials
 from oauth2client.service_account import ServiceAccountCredentials
 
@@ -44,6 +46,8 @@ class GCECloudProvider(BaseCloudProvider):
         self.region_name = self._get_config_value(
             'gce_region_name', 'us-central1')
 
+        # oauth2client.Credentials to be used for authentication
+        self._credentials = None
         # service connections, lazily initialized
         self._gce_compute = None
 
@@ -86,7 +90,26 @@ class GCECloudProvider(BaseCloudProvider):
                 self.credentials_dict)
         else:
             credentials = GoogleCredentials.get_application_default()
+        self._credentials = credentials
         return discovery.build('compute', 'v1', credentials=credentials)
+
+    def get_gce_resource_data(self, uri):
+        """
+        Retrieves GCE resoure data given its resource URI.
+        """
+        http = httplib2.Http()
+        http = self._credentials.authorize(http)
+        def _postproc(*kwargs):
+            if len(kwargs) >= 2:
+                # The first argument is request, and the second is response.
+                resource_dict = json.loads(kwargs[1])
+                return resource_dict
+        request = googleapiclient.http.HttpRequest(http=http,
+                                                   postproc=_postproc,
+                                                   uri=uri)
+        # The response is a dict representing the GCE resource data.
+        response = request.execute()
+        return response
 
     def wait_for_global_operation(self, operation):
         while True:
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -746,7 +746,8 @@ class GCEInstance(BaseInstance):
         # In GCE, the name of the instance is provided by the client when
         # initially creating the resource. The name cannot be changed after
         # the instance is created.
-        raise NotImplementedError("Not implemented by this provider.")
+        cb.log.warning("Setting instance name after it is created is not "
+                       "supported by this provider.")
 
     @property
     def public_ips(self):
@@ -871,19 +872,18 @@ class GCEInstance(BaseInstance):
         """
         Get the security groups associated with this instance.
         """
-        sg_service = self._provider.security.security_groups
-        sg_names = [sg.name for sg in sg_service.list()]
-        tags = self._gce_instance['tags']['items']
-        if len(tags) == 0 or len(sg_names) == 0:
+        network_url = self._gce_instance.get('networkInterfaces')[0].get(
+            'network')
+        network_name = GCEFirewallsDelegate.network_name(
+            {'network': network_url})
+        if 'items' not in self._gce_instance['tags']:
             return []
-        common_tags = set(sg_names) & set(tags)
-        # We have a list of the security group names, now get the
-        # GCESecurityGroup objects.
-        sgs = []
-        for tag in common_tags:
-            result = sg_service.find(tag)
-            if len(result) > 0:
-                sgs.append(result[0])
+        tags = self._gce_instance['tags']['items']
+        # Tags are mapped to non-empty security groups under the instance
+        # network. Unmatched tags are ignored.
+        sgs = (self._provider.security
+               .security_groups.find_by_network_and_tags(
+                   network_name, tags))
         return sgs
 
     @property
@@ -942,6 +942,10 @@ class GCENetwork(BaseNetwork):
     def __init__(self, provider, network):
         super(GCENetwork, self).__init__(provider)
         self._network = network
+
+    @property
+    def resource_url(self):
+        return self._network['selfLink']
 
     @property
     def id(self):
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -1,6 +1,7 @@
 """
 DataTypes used by this provider
 """
+from cloudbridge.cloud.base.resources import BaseInstance
 from cloudbridge.cloud.base.resources import BaseInstanceType
 from cloudbridge.cloud.base.resources import BaseKeyPair
 from cloudbridge.cloud.base.resources import BaseMachineImage
@@ -8,6 +9,8 @@ from cloudbridge.cloud.base.resources import BasePlacementZone
 from cloudbridge.cloud.base.resources import BaseRegion
 from cloudbridge.cloud.base.resources import BaseSecurityGroup
 from cloudbridge.cloud.base.resources import BaseSecurityGroupRule
+from cloudbridge.cloud.interfaces.resources import InstanceState
+from cloudbridge.cloud.interfaces.resources import MachineImageState
 
 # Older versions of Python do not have a built-in set data-structure.
 try:
@@ -64,6 +67,10 @@ class GCEInstanceType(BaseInstanceType):
     def __init__(self, provider, instance_dict):
         super(GCEInstanceType, self).__init__(provider)
         self._inst_dict = instance_dict
+
+    @property
+    def resource_url(self):
+        return self._inst_dict.get('selfLink')
 
     @property
     def id(self):
@@ -615,6 +622,10 @@ class GCEMachineImage(BaseMachineImage):
             self._gce_image = image
 
     @property
+    def resource_url(self):
+        return self._gce_image.get('selfLink')
+
+    @property
     def id(self):
         """
         Get the image identifier.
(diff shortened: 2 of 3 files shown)


cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -690,6 +700,244 @@ def refresh(self):
...
+            self._provider.gce_compute \
+                          .instances() \
+                          .reset(project=self._provider.project_name,
+                                 zone=self._provider.default_zone,
+                                 instance=self.name) \
+                          .execute()
+
+    def terminate(self):
+        """
+        Permanently terminate this instance.
```
mbookman: This does not "permanently" terminate the instance, since you can restart it

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -690,6 +700,244 @@ def refresh(self):
...
+        """
+        Get the security groups associated with this instance.
+        """
+        sg_service = self._provider.security.security_groups
+        sg_names = [sg.name for sg in sg_service.list()]
+        tags = self._gce_instance['tags']['items']
+        if len(tags) == 0 or len(sg_names) == 0:
+            return []
+        common_tags = set(sg_names) & set(tags)
+        sgs = []
```
mbookman: Can you add a comment here:

    # We have a list of the security group names, now get the GCESecurityGroup objects.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -402,10 +405,133 @@ def list(self, limit=None, marker=None):
...
+        Creates a new virtual machine instance.
+        """
+        if zone is None:
+            zone = self.provider.default_zone
+        if launch_config is None:
+            config = {
+                'name': name,
+                'machineType': instance_type.resource_url,
+                'disks': [{'boot': True,
+                           'autoDelete': False,
```
mbookman: Why is autoDelete set to false?
One typically expects the boot disk to be deleted along with the instance.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -402,10 +405,133 @@ def list(self, limit=None, marker=None):
...
+                           }
+                       }],
+                'networkInterfaces': [
+                    {'network': 'global/networks/default',
+                     'accessConfigs': [{'type': 'ONE_TO_ONE_NAT',
+                                        'name': 'External NAT'}]
+                 }],
+            }
+            if (security_groups is not None and
+                isinstance(security_groups, list) and
```
mbookman: What else would security_groups be, besides a list? Should that case be handled?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -402,10 +405,133 @@ def list(self, limit=None, marker=None):
...
+                len(security_groups) > 0):
+                sg_names = []
+                if isinstance(security_groups[0], SecurityGroup):
+                    sg_namess = [sg.name for sg in security_groups]
```
mbookman: s/sg_namess/sg_names

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -402,10 +405,133 @@ def list(self, limit=None, marker=None):
...
+    def create(self, name, image, instance_type, zone=None,
+               key_pair=None, security_groups=None, user_data=None,
+               launch_config=None,
+               **kwargs):
+        """
+        Creates a new virtual machine instance.
+        """
+        if zone is None:
+            zone = self.provider.default_zone
+        if launch_config is None:
```
mbookman: [Style guide](https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations) suggests using implicit True/False.

So this should be:

    if not launch_config:

Similar comment on the "if zone is None" check above. But I would defer to the project style convention.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -402,10 +405,133 @@ def list(self, limit=None, marker=None):
...
+                "googleapiclient.errors.HttpError: {0}".format(http_error))
+        return None
+
+    def find(self, name, limit=None, marker=None):
+        """
+        Searches for instances by instance name.
+        :return: a list of Instance objects
+        """
+        instances = []
+        for instance in self.list():
```
mbookman: More pythonic would be:

    instances = [instance.name for instance in self.list() if instance.name == name]

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -402,10 +405,133 @@ def list(self, limit=None, marker=None):
...
+        if limit and len(instances) > limit:
+            instances = instances[:limit]
+        return instances
+
+    def list(self, limit=None, marker=None):
+        """
+        List all instances.
+        """
+        if limit is None or limit > 500:
+            # For GCE API, Acceptable values are 0 to 500, inclusive.
```
mbookman: Can you put this comment ahead of the previous line so that the code itself is compact?

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -5,11 +5,13 @@
 
 
 from cloudbridge.cloud.base import BaseCloudProvider
+import httplib2
```
baizhang: Thanks for pointing this out. The discussion in that thread was very interesting. For this particular GCEInstanceService implementation, we actually create a new Http() object for each request, as suggested here https://developers.google.com/api-client-library/python/guide/thread_safety

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -690,6 +700,244 @@ def refresh(self):
...
+    @property
+    def name(self):
+        """
+        Get the instance name.
+        """
+        return self._gce_instance['name']
+
+    @name.setter
+    # pylint:disable=arguments-differ
+    def name(self, value):
```
baizhang: Done. I added a comment here that "In GCE, the name of the instance is provided by the client when initially creating the resource. The name cannot be changed after the instance is created", and it will raise NotImplementedError("Not implemented by this provider.")

cloudbridge/cloud/providers/gce/resources.py
baizhang: Done. I changed the terminate method to actually delete the instance. I think the GCE status 'TERMINATED' should be mapped to InstanceState.STOPPED, if I understand correctly. I also added a stop method, which terminates (in GCE parlance) an instance.

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/services.py

cloudbridge/cloud/providers/gce/services.py

cloudbridge/cloud/providers/gce/services.py
baizhang: This roughly follows the same logic as implemented for AWS.

https://github.com/gvlproject/cloudbridge/blob/master/cloudbridge/cloud/providers/aws/services.py#L607

cloudbridge/cloud/providers/gce/services.py

cloudbridge/cloud/providers/gce/services.py

cloudbridge/cloud/providers/gce/services.py

baizhang:
Hi @mbookman , thanks for your review. I updated this pull request. Please take another look. Thanks!

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -680,4 +691,243 @@ def refresh(self):
...
+        tags = self._gce_instance['tags']['items']
+        if len(tags) == 0 or len(sg_names) == 0:
+            return []
+        common_tags = set(sg_names) & set(tags)
+        sgs = []
+        for tag in common_tags:
+            result = sg_service.find(tag)
+            if len(result) > 0:
+                sgs.append(result[0])
+        return sgs
```
chiniforooshan: Shouldn't we filter out SGs that are not in the same network as the instance?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -680,4 +691,243 @@ def refresh(self):
...
+        zone_uri = self._gce_instance.get('zone')
+        if zone_uri is None:
+            return None
+        zone = self._provider.get_gce_resource_data(zone_uri)
+        return zone.get('name', None)
+
+    @property
+    def security_groups(self):
+        """
+        Get the security groups associated with this instance.
```
chiniforooshan: Sorry, I deleted my comment by mistake! Deleted comment was asking whether we should return non-empty security groups here (i.e. current implementation) or all security groups (i.e. by just looking at instance tags) to match the security group given at the creation time of the instance?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -394,10 +397,133 @@ def list(self, limit=None, marker=None):
                                      limit=limit, marker=marker)
 
 
+class GCEInstanceService(BaseInstanceService):
+
+    def __init__(self, provider):
+        super(GCEInstanceService, self).__init__(provider)
+
+    def create(self, name, image, instance_type, zone=None,
```
chiniforooshan: There is a "network" parameter in the create method of the InstanceService interface, after instance_type, before zone.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -435,12 +435,12 @@ def create(self, name, image, instance_type, zone=None,
                                         'name': 'External NAT'}]
                  }],
             }
-            if (security_groups is not None and
+            if (security_groups and
                 isinstance(security_groups, list) and
                 len(security_groups) > 0):
```
mbookman: I think that checking len(security_groups) isn't needed here.
If security_groups is an empty list then "if security_groups" returns False.

cloudbridge/cloud/providers/gce/resources.py
nuwang: @baizhang @mbookman @chiniforooshan @afgane I'm wondering whether we should consider silently ignoring this? On the one hand, setting the name does not appear to be a critical function to me , and if that rather innocuous action requires a try/exception block, that would seem to complicate the client interface and/or make it difficult to write cross-cloud code? On the other hand, there genuinely is a difference in functionality, and attempting to gloss over it may lull the user into a false sense of having succeeded. Perhaps the compromise is to document, at the interface level, that setting the name is not guaranteed to succeed, and code should not rely on the name being set?

baizhang: I changed NotImplementedError exception to cb.log.warning. I agree that we may also need to document this at the interface level.

cloudbridge/cloud/providers/gce/resources.py
baizhang: Done. I added a method find_by_network_and_tags to GCESecurityGroupService to map tags to security groups by network and tags.

cloudbridge/cloud/providers/gce/resources.py
baizhang: Done. I added the comment that "tags are mapped to non-empty security groups under the instance network. Unmatched tags are ignored".

cloudbridge/cloud/providers/gce/services.py

cloudbridge/cloud/providers/gce/services.py

baizhang:
Hi @chiniforooshan @mbookman @nuwang , thank you for your review comments.

Please take another look. Thanks!

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -410,16 +427,20 @@ class GCEInstanceService(BaseInstanceService):
...
-               **kwargs):
+               launch_config=None, **kwargs):
         """
         Creates a new virtual machine instance.
         """
         if not zone:
             zone = self.provider.default_zone
         if not launch_config:
+            if network:
+                network_url = (network.resource_url
```
chiniforooshan: I think these short forms are good when you can fit the whole thing in one line; otherwise, a complete if statement is cleaner, IMHO. Feel free to ignore, specially since it looks like there are statements like this in other parts of the code.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -402,10 +422,133 @@ def list(self, limit=None, marker=None):
...
+            config = launch_config
+        response = self.provider.gce_compute \
+                                .instances() \
+                                .insert(
+                                    project=self.provider.project_name,
+                                    zone=self.provider.default_zone,
+                                    body=config) \
+                                .execute()
+
+    def get(self, instance_name):
```
nuwang: @baizhang @mbookman Just noticed that there's an interface mismatch issue here: the get method should find an instance by id, not by name. I did see your comment that the GCE APIs do not allow finding an instance by id. So does that mean the GCE APIs use the instance_name as the primary key? How can you unambiguously retrieve a particular instance in GCE? From a CloudBridge perspective, I think the following equality should hold true:
instance == provider.instances.get(instance.id)
Therefore, whatever the instance.id property returns needs to be a unique identifier for an instance, be it a string, a UUID or something else.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -402,10 +422,133 @@ def list(self, limit=None, marker=None):
...
+
+    def list(self, limit=None, marker=None):
+        """
+        List all instances.
+        """
+        # For GCE API, Acceptable values are 0 to 500, inclusive.
+        # (Default: 500). If the number of available results is larger
+        # than maxResults, Compute Engine returns a nextPageToken that
+        # can be used to get the next page of results in subsequent
+        # list requests.
```
nuwang: @baizhang I realised that the interface does not adequately capture how the limit should be dealt with. I think that the implementation should not forced to honour the limit requested by the user. It can be a lower limit. This way, the implementation can be made much simpler. I've also added a commit which clarifies this contract https://github.com/gvlproject/cloudbridge/commit/919069f75a00a063513e64fc9a0538ef2094515f

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -402,10 +422,133 @@ def list(self, limit=None, marker=None):
...
+                              for inst in response['items']])
+            if limit and len(instances) >= limit:
+                break
+            request = self.provider.gce_compute \
+                                   .instances() \
+                                   .list_next(previous_request=request,
+                                              previous_response=response)
+        if limit and len(instances) > limit:
+            instances = instances[:limit]
+        return instances
```
nuwang: @baizhang There's a small interface mismatch here. The list method should return a ResultList object, instead of a List object. In this case, this can probably be wrapped in a ServerPagedResultList object.

Combined with the comment above, I think this function could possibly be rewritten as:
```
        max_result = limit if limit < 500 else 500
        response = self.provider.gce_compute.instances().list(project=self.provider.project_name,
                                     zone=self.provider.default_zone,
                                     maxResults=max_result,
                                     pageToken=marker)
        instances = [GCEInstance(self.provider, inst)
                             for inst in response['items']]
        return ServerPagedResultList(len(instances) > max_result,
                                     response['nextPageToken'],
                                     False, data=instances)
```

cloudbridge/cloud/providers/gce/services.py
mbookman: GCE instances are uniquely identified by project/zone/instance-name:

https://cloud.google.com/compute/docs/reference/latest/instances/get

So while the GCE API does assign a unique ID, I don't believe that it is useful and the interface here should use the [selfLink](https://cloud.google.com/compute/docs/reference/latest/instances#resource) as the id.

nuwang: @mbookman Sounds like the selfLink would work well. The ID property would be rarely, if ever, displayed to a user.

cloudbridge/cloud/providers/gce/services.py

cloudbridge/cloud/providers/gce/services.py

baizhang:
Hi @nuwang ! Thank you for your review.

Please take another look at the updated pull request. Thanks!

nuwang:
Thanks @baizhang, looks great to me. There are some conflicts with the current state of the branch, and I'm unsure how to merge them (provider.py in particular), so would you be able to update the pull request?

baizhang:
Hi @nuwang ! I merged the changes, and manually tested GCEInstanceService.

Please take another look. Thanks!

nuwang:
@baizhang Thanks!

PR #28, opened 2017-02-02
New Features: (1) Upload a file to bucket, (2) list keys with a prefix, (3) check for the existence of key within bucket, and (4) get a URL (with TTL) for a given object. 

VJalili:
1. New function `upload_from_file` is added. Accordingly, interfaces and `AWSBucketObject` are updated. The signature of the function is added to `OpenStackBucketObject`, but not implemented.
2. Test codes for AWS `upload_from_file` added.

### DIFF
--- a/cloudbridge/cloud/providers/aws/resources.py
+++ b/cloudbridge/cloud/providers/aws/resources.py
@@ -38,7 +38,6 @@ from retrying import retry
 
 
 class AWSMachineImage(BaseMachineImage):
-
     IMAGE_STATE_MAP = {
         'pending': MachineImageState.PENDING,
         'available': MachineImageState.AVAILABLE,
@@ -109,7 +108,6 @@ class AWSMachineImage(BaseMachineImage):
 
 
 class AWSPlacementZone(BasePlacementZone):
-
     def __init__(self, provider, zone, region):
         super(AWSPlacementZone, self).__init__(provider)
         if isinstance(zone, AWSPlacementZone):
@@ -152,7 +150,6 @@ class AWSPlacementZone(BasePlacementZone):
 
 
 class AWSInstanceType(BaseInstanceType):
-
     def __init__(self, provider, instance_dict):
         super(AWSInstanceType, self).__init__(provider)
         self._inst_dict = instance_dict
@@ -204,7 +201,6 @@ class AWSInstanceType(BaseInstanceType):
 
 
 class AWSInstance(BaseInstance):
-
     # ref:
     # http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html
     INSTANCE_STATE_MAP = {
@@ -373,7 +369,6 @@ class AWSInstance(BaseInstance):
 
 
 class AWSVolume(BaseVolume):
-
     # Ref:
     # http://docs.aws.amazon.com/AWSEC2/latest/CommandLineReference/
     # ApiReference-cmd-DescribeVolumes.html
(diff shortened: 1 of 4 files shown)


nuwang:
@VJalili Thanks for this pull request. We won't be able to get to this for a few more days because both Enis and I are at the same conference. We'll get back to you soon. Just a preliminary comment - I think that "upload_from_large_file" needs to be better defined. What would constitute a large file? Also, both upload functions accept a path, but what about a file stream? I think it would be great to make a comparison with other libraries like libcloud or boto to better define the interface.

VJalili:
@nuwang 
1. It seems the maximum file size to be uploaded using `set_contents_from_filename` is **5GB**:[https://groups.google.com/forum/#!topic/boto-users/CM1eZFgk-JA](url)
Given that we might regularly upload files larger than **5GB** (in Galaxy), so, it would be great if we could implement this functionality in CloudBridge too. 

2. `boto` also accepts filenames; it would be a good idea to have both signatures of the functions, one with `file` and other with `string`. I chose `string` (as first to try) mainly because _Galaxy Objectstore_ is using `string`.

cloudbridge/cloud/providers/aws/resources.py
```diff
@@ -858,19 +856,24 @@ def get(self, key):
...
+    def list(self, limit=None, marker=None, prefix=None):
         """
         List all objects within this bucket.
 
         :rtype: BucketObject
         :return: List of all available BucketObjects within this bucket.
         """
-        objects = [AWSBucketObject(self._provider, obj)
-                   for obj in self._bucket.list()]
+        if prefix:
```
afgane: I think you can omit this `if` here and just pass `prefix` regardless because if it's `None` it will just default to boto's default value anyhow.

cloudbridge/cloud/providers/aws/resources.py
```diff
@@ -858,19 +856,24 @@ def get(self, key):
         Retrieve a given object from this bucket.
         """
         key = Key(self._bucket, key)
-        if key.exists():
+        if key and key.exists():
             return AWSBucketObject(self._provider, key)
         return None
 
-    def list(self, limit=None, marker=None):
+    def list(self, limit=None, marker=None, prefix=None):
```
afgane: The `prefix` parameter does not exist in the interface.

cloudbridge/cloud/providers/aws/resources.py
```diff
@@ -38,7 +38,6 @@
 
 
 class AWSMachineImage(BaseMachineImage):
-
```
afgane: All of these extra lines are per PEP8 so they should remain: https://www.python.org/dev/peps/pep-0008/#blank-lines

cloudbridge/cloud/providers/aws/resources.py
```diff
@@ -703,7 +694,7 @@ def id(self):
         md5 = hashlib.md5()
         md5.update("{0}-{1}-{2}-{3}".format(
             self.ip_protocol, self.from_port, self.to_port, self.cidr_ip)
-            .encode('ascii'))
```
afgane: This also looks like it was using proper PEP8 formatting and will now trigger a hanging indent complaint: https://www.python.org/dev/peps/pep-0008/#id3

I'd suggest installing a plugin like flake8 into the editor to do these checks automatically (or manually running it): http://flake8.pycqa.org/en/latest/

cloudbridge/cloud/providers/aws/resources.py
VJalili: This commit was missing; updated.

afgane:
Merging this now but will follow up with a few minor changes.

PR #29, opened 2017-02-07
NetworkService: Floating IPs

chiniforooshan:
This implements the floating IP resource using GCP's static IPs and forwarding rules. Please take a look.

@baizhang, @mbookman, @nuwang, @afgane

### DIFF
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -37,14 +37,42 @@ class GCPResources(object):
     def __init__(self, connection):
         self._connection = connection
 
-        # Sorry, but the most reliable source of resource descriptions is the
-        # internal _resourceDesc field of the connection.
+        # Resource descriptions are already pulled into the internal
+        # _resourceDesc field of the connection.
         #
         # TODO: We could fetch compute resource descriptions from
         # https://www.googleapis.com/discovery/v1/apis/compute/v1/rest and
         # storage resource descriptions from
         # https://www.googleapis.com/discovery/v1/apis/storage/v1/rest
         # ourselves.
+        #
+        # Resource descriptions are in JSON format which are then parsed into a
+        # Python dictionary. The main fields we are interested are:
+        #
+        # {
+        #   "rootUrl": "https://www.googleapis.com/",
+        #   "servicePath": COMPUTE OR STORAGE SERVICE PATH
+        #   "resources": {
+        #     RESOURCE_NAME: {
+        #       "methods": {
+        #         "get": {
+        #           "path": RESOURCE PATH PATTERN
+        #           "parameters": {
+        #             PARAMETER: {
+        #               "pattern": REGEXP FOR VALID VALUES
+        #               ...
+        #             },
+        #             ...
+        #           },
+        #           "parameterOrder": [LIST OF PARAMETERS]
+        #         },
+        #         ...
+        #       }
+        #     },
+        #     ...
+        #   }
+        #   ...
+        # }
         desc = connection._resourceDesc
         self._root_url = desc['rootUrl']
         self._service_path = desc['servicePath']
@@ -76,6 +104,11 @@ class GCPResources(object):
                                          'pattern': re.compile(pattern)}
 
     def parse_url(self, url):
+        """
+        Build a GCPResourceUrl from a resource's URL string. One can then call
+        the get() method on the returned object to fetch resource details from
+        GCP servers.
+        """
         url = url.strip()
         if url.startswith(self._root_url): url = url[len(self._root_url):]
         if url.startswith(self._service_path):
@@ -128,8 +161,7 @@ class GCECloudProvider(BaseCloudProvider):
         self._network = GCENetworkService(self)
 
         self._compute_resources = GCPResources(self.gce_compute)
-        # Enable when the storage 
-        # self._storage_resources = GCPResources(self.gcp_storage)
+        self._storage_resources = GCPResources(self.gcp_storage)
 
     @property
     def compute(self):
@@ -161,7 +193,7 @@ class GCECloudProvider(BaseCloudProvider):
 
     @property
     def gcp_storage(self):
-        if not self._gcp_storae:
+        if not self._gcp_storage:
             self._gcp_storage = self._connect_gcp_storage()
         return self._gcp_storage
 
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -5,9 +5,10 @@ for GCE.
 
 
 from cloudbridge.cloud.base import BaseCloudProvider
-import cloudbridge as cb
 import json
 import os
+import re
+from string import Template
 import time
 
 from googleapiclient import discovery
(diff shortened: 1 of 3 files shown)


cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -23,6 +26,95 @@
...
+        out.name = parts[-1]
+
+        # If a cloud provider is given, set the connection.
+        if provider:
+            if out.platform == 'compute':
+                out.connection = provider.gce_compute
+            elif out.platform == 'storage':
+                out.connection = provider.gcp_storage
+
+        # Set region, zone, and project fields.
```
mbookman: Please add a comment regarding what you are looking for here in the URL.
Again, an example goes a long way.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -739,3 +826,72 @@ def create_subnet(self, cidr_block, name=None):
...
+        # global IPs can be forwarded only to load balancing resources, not to
+        # a specific instance. Find out the region to which the IP belongs.
+        self._region = GCPResourceUrl.parse_url(self._ip['region']).name
+
+        # Check if the address is used by a resource.
+        self._rule = None
+        self._target_instance = None
+        if 'users' in floating_ip and len(floating_ip['users']) > 0:
+            if len(floating_ip['users']) > 1:
+                cb.log.warning('IP is in user by more than one resource')
```
mbookman: cb.log.warning('Address "%s" in use by more than one resource', floating_ip['address'])
Similar change to format "Address ..." to the two warnings below.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -23,6 +26,95 @@
...
+        self.platform = platform
+        self.resource_type = resource_type
+
+    @property
+    def kind(self):
+        return '%s#%s' % (self.platform, self.resource_type)
+
+    @property
+    def url_field_name(self):
+        return '%ss' % self.resource_type
```
chiniforooshan: Since it was not too difficult, I decided to directly use the resource descriptions coming from googleapis.com instead of using heuristics like this. Please take another look at the new implementation.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -23,6 +26,95 @@
...
+
+
+# GCP resource types we are interested in
+FORWARDING_RULE_TYPE = GCPResourceType('compute', 'forwardingRule')
+INSTANCE_TYPE = GCPResourceType('compute', 'instance')
+NETWORK_TYPE = GCPResourceType('compute', 'network')
+REGION_TYPE = GCPResourceType('compute', 'region')
+TARGET_INSTANCE_TYPE = GCPResourceType('compute', 'targetInstance')
+
```
chiniforooshan: The new implementation does not need this. It gets the list of all available resources from googleapis.com (via the discovery library).

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -23,6 +26,95 @@
...
+        super(GCPResourceUrl, self).__init__(platform, resource_type)
+        self.project = project
+        self.region = region
+        self.zone = zone
+        self.name = name
+        self.connection = connection
+
+    @staticmethod
+    def parse_url(url, provider=None):
+        parts = url.split('/')
```
chiniforooshan: Removed all these heuristics in the new implementation.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -23,6 +26,95 @@
...
+        out = None
+        for resource_type in ALL_RESOURCE_TYPES:
+            if parts[-2] == resource_type.url_field_name:
+                out = GCPResourceUrl(resource_type.platform,
+                                     resource_type.resource_type)
+        if not out:
+            return out
```
chiniforooshan: Re-implemented.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Re-implemented.

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -19,6 +20,77 @@
...
+        discovery_object = getattr(self._connection, self._resource)()
+        return discovery_object.get(**self.parameters).execute()
+
+
+class GCPResources(object):
+    
+    def __init__(self, connection):
+        self._connection = connection
+
+        # Sorry, but the most reliable source of resource descriptions is the
```
mbookman: I don't think you need to apologize.
Can you provide a short example of what the content looks like (or is there a document/example online we could link to)?

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -19,6 +20,77 @@
...
+                    mapping[parameter] = '(%s)' % parameter_desc['pattern']
+                else:
+                    mapping[parameter] = '([^/]+)'
+            pattern = template.substitute(**mapping)
+
+            # Store the parameters and the regex pattern of this resource.
+            self._resources[resource] = {'parameters': parameters,
+                                         'pattern': re.compile(pattern)}
+
+    def parse_url(self, url):
```
mbookman: I like how compact this class is. I think it's compactness makes it a bit obscure.
Please add some comments regarding expecting inputs and outputs.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -48,12 +120,17 @@ def __init__(self, config):
...
         self._gce_compute = None
+        self._gcp_storage = None
 
         # Initialize provider services
         self._compute = GCEComputeService(self)
         self._security = GCESecurityService(self)
         self._network = GCENetworkService(self)
 
+        self._compute_resources = GCPResources(self.gce_compute)
+        # Enable when the storage
```
mbookman: This comment appears incomplete.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -84,15 +161,23 @@ def gce_compute(self):
 
     @property
     def gcp_storage(self):
-        raise NotImplementedError("To be implemented")
+        if not self._gcp_storae:
```
mbookman: s/_gcp_storae/_gcp_storage

cloudbridge/cloud/providers/gce/provider.py
chiniforooshan: sdf

chiniforooshan: Didn't find an online document about the format. Added some explanation here.

cloudbridge/cloud/providers/gce/provider.py
chiniforooshan: Done.

cloudbridge/cloud/providers/gce/provider.py
chiniforooshan: Done.

cloudbridge/cloud/providers/gce/provider.py
chiniforooshan: Done.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -489,11 +493,45 @@ def create(self, name):
...
+                                     .insert(project=self.provider.project_name,
+                                             region=region,
+                                             body={'name': ip_name})
+                                     .execute())
+            if 'error' in response:
+                return None
+            self.provider.wait_for_operation(response, region=region)
+            ips = self.floating_ips()
+            for ip in ips:
+                if ip.id == response["targetId"]:
```
nuwang: @afgane This looks like the equivalent of a FloatingIP.get() method, but we don't have a dedicated floating IP service. Should we?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -739,3 +738,74 @@ def create_subnet(self, cidr_block, name=None):
...
+    def id(self):
+        return self._ip['id']
+
+    @property
+    def public_ip(self):
+        return self._ip['address']
+
+    @property
+    def private_ip(self):
+        if not self._target_instance:
```
nuwang: @chiniforooshan There is a reference to self._target_instance, but self._target_instance is initialized conditionally in the constructor and therefore, the property is not guaranteed to be available. Could this cause a problem?

cloudbridge/cloud/providers/gce/services.py
afgane: I don't think there's a real need for it. From the design perspective, sure. However, the functionality has small scope so rather than bloating the library, this is probably sufficient. If we add support for load balancers and other possible services that can be associated with an IP, then it may become better separated to do so but in the mean time, I vote to just leave it as part of the `NetworkService`.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: It should be OK. In the constructor, self._targe_instance is set to None by default. If this IP is assigned to an instance, then self._target_instance is set to that instance. So, throughout this code, self._target_instance == None means that this IP is not assigned to an instance (e.g. see in_use() implementation), and should be handled correctly.

nuwang: @chiniforooshan Sorry, I missed that line. All good.

PR #31, opened 2017-03-15
add/remove floating IPs to instances

chiniforooshan:
@baizhang, @mbookman, @nuwang, @afgane

### DIFF
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -82,6 +82,8 @@ class GCPResources(object):
 
         # We will not mutate self._desc; it's OK to use items() in Python 2.x.
         for resource, resource_desc in desc['resources'].items():
+            if 'methods' not in resource_desc:
+                continue
             methods = resource_desc['methods']
             if 'get' not in methods:
                 continue
@@ -163,7 +165,7 @@ class GCECloudProvider(BaseCloudProvider):
         self._network = GCENetworkService(self)
 
         self._compute_resources = GCPResources(self.gce_compute)
-        self._storage_resources = GCPResources(self.gcp_storage)
+        # self._storage_resources = GCPResources(self.gcp_storage)
 
     @property
     def compute(self):
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -82,10 +82,8 @@ class GCPResources(object):
 
         # We will not mutate self._desc; it's OK to use items() in Python 2.x.
         for resource, resource_desc in desc['resources'].items():
-            if 'methods' not in resource_desc:
-                continue
-            methods = resource_desc['methods']
-            if 'get' not in methods:
+            methods = resource_desc.get('methods', {})
+            if not methods.get('get'):
                 continue
             method = methods['get']
             parameters = method['parameterOrder']
@@ -165,7 +163,7 @@ class GCECloudProvider(BaseCloudProvider):
         self._network = GCENetworkService(self)
 
         self._compute_resources = GCPResources(self.gce_compute)
-        # self._storage_resources = GCPResources(self.gcp_storage)
+        self._storage_resources = GCPResources(self.gcp_storage)
 
     @property
     def compute(self):
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -912,19 +912,193 @@ class GCEInstance(BaseInstance):
         raise NotImplementedError(
             'To be implemented after GCEVolumeService.')
 
+    @property
+    def _existing_target_instance(self):
+        parsed_url = self._provider.parse_url(self._gce_instance['selfLink'])
+        try:
+            response = (self._provider.gce_compute
+                                      .targetInstances()
+                                      .list(project=self._provider.project_name,
+                                            zone=parsed_url.parameters['zone'])
+                                      .execute())
+            if 'items' not in response:
+                return None
+            for target_instance in response['items']:
+                instance = self._provider.parse_url(target_instance['instance'])
+                if instance.parameters['instance'] == self.name:
+                    return target_instance
+        except Exception as e:
+            cb.log.warning('Exception while listing target instances: %s', e)
+        return None
+
+    @property
+    def _target_instance(self):
+        existing_target_instance = self._existing_target_instance
+        if existing_target_instance:
+            return existing_target_instance
+
+        # No targetInstance exists for this instance. Create one.
+        url = self._gce_instance['selfLink']
+        parsed_url = self._provider.parse_url(url)
+        body = {'name': 'target-instance-{0}'.format(uuid.uuid4()),
+                'instance': url}
+        try:
+            response = (self._provider.gce_compute
+                                      .targetInstances()
+                                      .insert(
+                                          project=self._provider.project_name,
+                                          zone=parsed_url.parameters['zone'],
+                                          body=body)
+                                      .execute())
+            self._provider.wait_for_operation(
+                response, zone=parsed_url.parameters['zone'])
+        except Exception as e:
+            cb.log.warning('Exception while inserting a target instance: %s', e)
+            return None
+
+        return self._existing_target_instance
+
+    def _redirect_existing_rule(self, ip, target_instance):
+        new_zone = (self._provider.parse_url(target_instance['zone'])
+                                  .parameters['zone'])
+        new_name = target_instance['name']
+        new_url = target_instance['selfLink']
+        try:
+            response = (self._provider.gce_compute
+                                      .forwardingRules()
+                                      .list(project=self._provider.project_name,
+                                            region=ip.region)
+                                      .execute())
+            if 'items' not in response:
+                return False
+            for rule in response['items']:
+                if rule['IPAddress'] == ip.public_ip:
+                    parsed_target_url = self._provider.parse_url(rule['target'])
+                    old_zone = parsed_target_url.parameters['zone']
+                    old_name = parsed_target_url.parameters['targetInstance']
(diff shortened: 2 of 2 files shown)


cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +912,191 @@ def create_image(self, name):
         raise NotImplementedError(
             'To be implemented after GCEVolumeService.')
 
+    @property
+    def _existing_target_instance(self):
```
afgane: I feel all the properties here, even though they're intended as private, should have a docstring included to make it clear right away what's their purpose.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +912,191 @@ def create_image(self, name):
...
+                    return
+                target_instance = self._target_instance
+                if not target_instance:
+                    # We should not be here.
+                    cb.log.warning('Something went wrong! "%s" is associated '
+                                   'to "%s" with no target instance',
+                                   ip_address, self.name)
+                    return
+                if not self._delete_existing_rule(ip, target_instance):
+                    cb.log.warning('Could not remove "%s" from isntance "%s"',
```
afgane: Typo here for `instances`. Also, this could probably be a bit more explicit: `Cloud not remove floating IP ...`

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +912,191 @@ def create_image(self, name):
         raise NotImplementedError(
             'To be implemented after GCEVolumeService.')
 
+    @property
+    def _existing_target_instance(self):
+        self_url = self._provider.parse_url(self._gce_instance['selfLink'])
+        try:
+            response = (self._provider.gce_compute
+                                      .targetInstances()
+                                      .list(project=self._provider.project_name,
```
chiniforooshan: Yes, it is a part of the selfLink. But, is there a reason to not use `self._provider.project_name`? It is consistent with all other places where we use the `list()` method, and less importantly, it is 3 characters shorter than `self_url.parameters['project']`, which makes it fit in the same line :).

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +912,191 @@ def create_image(self, name):
...
+                url = self._provider.parse_url(target_instance['instance'])
+                if url.parameters['instance'] == self.name:
+                    return target_instance
+        except Exception as e:
+            cb.log.warning('Exception while listing target instances: %s', e)
+        return None
+
+    @property
+    def _target_instance(self):
+        existing_target_instance = self._existing_target_instance
```
chiniforooshan: It is not needed, but it caches the result of `self._existing_target_instance`, which may save an API call (note that `self._existing_target_instance` may try to list target instances).

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +912,191 @@ def create_image(self, name):
...
+
+        # No targetInstance exists for this instance. Create one.
+        self_url = self._provider.parse_url(self._gce_instance['selfLink'])
+        body = {'name': 'target-instance-{0}'.format(uuid.uuid4()),
+                'instance': self._gce_instance['selfLink']}
+        try:
+            response = (self._provider.gce_compute
+                                      .targetInstances()
+                                      .insert(
+                                          project=self._provider.project_name,
```
chiniforooshan: As explained above, if you are OK with using `self._provider.project_name` I prefer to be consistent. I think we have used `self._provider.project_name` in all API calls.

chiniforooshan: Acknowledged.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +912,191 @@ def create_image(self, name):
...
+                                          zone=self_url.parameters['zone'],
+                                          body=body)
+                                      .execute())
+            self._provider.wait_for_operation(
+                response, zone=self_url.parameters['zone'])
+        except Exception as e:
+            cb.log.warning('Exception while inserting a target instance: %s', e)
+            return None
+
+        return self._existing_target_instance
```

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +912,191 @@ def create_image(self, name):
...
+                                      .execute())
+            if 'items' not in response:
+                return False
+            for rule in response['items']:
+                if rule['IPAddress'] == ip.public_ip:
+                    parsed_target_url = self._provider.parse_url(rule['target'])
+                    old_zone = parsed_target_url.parameters['zone']
+                    old_name = parsed_target_url.parameters['targetInstance']
+                    if old_zone == new_zone and old_name == new_name:
+                        return
```
chiniforooshan: Good catch; done.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +912,191 @@ def create_image(self, name):
...
+                                    .forwardingRules()
+                                    .setTarget(
+                                        project=self._provider.project_name,
+                                        region=ip.region,
+                                        forwardingRule=rule['name'],
+                                        body={'target': new_url})
+                                    .execute())
+                    self._provider.wait_for_operation(response,
+                                                      region=ip.region)
+                    return True
```

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +912,191 @@ def create_image(self, name):
...
-            'To be implemented after GCENetworkService.')
+        for ip in self._provider.network.floating_ips():
+            if ip.public_ip == ip_address:
+                if ip.in_use():
+                    if ip.private_ip not in self.private_ips:
+                        cb.log.warning(
+                            'Floating IP "%s" is already associated to "%s".',
+                            ip_address, self.name)
+                    return
+                target_instance = self._target_instance
```
chiniforooshan: Calling `self._target_instance` several times may trigger several identical API calls, which are slow.

cloudbridge/cloud/providers/gce/resources.py

chiniforooshan:
The old implementation of the `public_ips` method just returns the NAT IP of the instance; it ignores forwarding rules. I fixed that too, in this PR.

cloudbridge/cloud/providers/gce/resources.py
mbookman: The one reason is that it is possible in a future version of things that the instance would be from a different Cloud project, so you'd want the fields to all come from the selfLink. Not a big deal today as everything is assumed to be in the same project.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -82,6 +82,8 @@ def __init__(self, connection):
 
         # We will not mutate self._desc; it's OK to use items() in Python 2.x.
         for resource, resource_desc in desc['resources'].items():
+            if 'methods' not in resource_desc:
```
mbookman: Don't know if this reads better but this could also be:

```
if not resource_desc.get('methods', {}).get('get'):
  continue
```

or

```
methods = resource_desc.get('methods', {})
if not methods.get('get'):
  continue
```

Also - it isn't clear why this block of code is here. What is the check for the "get" method for here? It is not referenced in the code here.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -163,7 +165,7 @@ def __init__(self, config):
         self._network = GCENetworkService(self)
 
         self._compute_resources = GCPResources(self.gce_compute)
-        self._storage_resources = GCPResources(self.gcp_storage)
+        # self._storage_resources = GCPResources(self.gcp_storage)
```
mbookman: Please add a comment regarding why this is commented out.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -912,19 +914,216 @@ def create_image(self, name):
         raise NotImplementedError(
             'To be implemented after GCEVolumeService.')
 
+    @property
```
mbookman: I'm wondering about the choice to make this a property, rather than an explicit function `get_target_instance(self)`. It feels like the labeling this a property leads to confusion when reading the code. "get_target_instance" or "lookup_target_instance" makes it more clear that there will be API calls involved.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Done

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Changed the name to a less confusing self._get_target_instance() name, as suggested by Matt.

cloudbridge/cloud/providers/gce/provider.py
chiniforooshan: Done.

cloudbridge/cloud/providers/gce/provider.py
chiniforooshan: It was a mistake :)

cloudbridge/cloud/providers/gce/resources.py

afgane:
Seems to be functional now so I'll merge it in and we can adjust if needed later:
```
>>> i.add_floating_ip(fip.public_ip)
>>> fip.public_ip
'104.197.88.254'
>>> gce.compute.instances.list()[0].public_ips
['35.184.237.197', '104.197.88.254']
```

One thing to note here is that GCE does not replace the default public IP with the one assigned to the instance but instead adds it. This is different from AWS, which replaces the ephemeral public IP with the floating one. Hence, if we made an assumption anywhere higher up in the stack of doing `i.public_ips[0]`, we'll probably need to revisit that code.

afgane:
Thanks for the continuous updates.

PR #33, opened 2017-04-10
Merge latest master into GCE + other fixes

nuwang:
This pull request does the following:
1. Replaces pycrypto with cryptography, as pycrypto is no longer maintained (commit: https://github.com/gvlproject/cloudbridge/commit/27b217e070e5711a7fd21c76663d64918fcd9c6e)
2. Applies the fix that Enis mentioned for getting GCPResources to work (https://github.com/gvlproject/cloudbridge/commit/42e6a221e202ffeafbd44185377a029d87d0148d)
3. Merges latest changes from master into the GCE branch (https://github.com/gvlproject/cloudbridge/commit/f06afe6647ef70a6b9912b22905fa198ac261bbb)
This includes changes to the network interface, where instance.create() now requires a subnet instead of a network and other fixes to simplify testing. With these changes, individual tests can be run when required, and tests are also skipped if not implemented. See testing section in docs for more info.
4. A preliminary patch or instance.create() to use subnet instead of network.
 https://github.com/gvlproject/cloudbridge/commit/10064bf047f26e73a7961837ffb61ca2fdec71a6


test/test_security_service.py
```diff
@@ -1,20 +1,19 @@
...
 class CloudSecurityServiceTestCase(ProviderTestBase):
 
-    def __init__(self, methodName, provider):
-        super(CloudSecurityServiceTestCase, self).__init__(
-            methodName=methodName, provider=provider)
-
+    @helpers.skipIfNoService(['security.key_pairs'])
     def test_crud_key_pair_service(self):
-        name = 'cbtestkeypairA-{0}'.format(uuid.uuid4()).lower()
+        name = 'cbtestkeypairA-{0}'.format(uuid.uuid4())
```
nuwang: Have removed this explicit .lower(), since the check is made by id.

nuwang:
@baizhang @chiniforooshan @mbookman What do you think about this pull request? I can merge the latest changes in again and update the pull request if it all looks ok. Let me know whether you are ok with the changes and whether to merge this in.

mbookman:
The one thing that jumps out to me is the subnet `zone` property.
GCE subnets are bound to a region and don't have the notion of "zone placement" as best as I can tell.
I don't think that this is a big deal. The zone will still be a parameter to instance creation, which is fine.

Would be great for @baizhang and @chiniforooshan to give it a look over.

baizhang:
LGTM. Thanks!

Issue #35, opened 2017-04-27
Doesn't upload files larger than 5 Gig to Swift


MartinPaulo:
If I use cloudbridge to upload a file larger than 5 Gig in size to Swift it eventually fails.

I've walked through the code and this is what I see happening: 
The `cloudbridge` `OpenStackBucketObject` `upload_from_file` method is called. This in turn eventually lands up calling the OpenStack `swiftclient.client.py` `Connection` `put_object` method: which is a low level method that does not attempt to break files > 5Gig in size into smaller segments. Hence when you hit the 5 Gig Swift wall it ends in tears..

The OpenStack utility class `SwiftService` has an upload method that will chunk files if you set the `segment_size` option. Thus this method will handle files > 5 Gig with aplomb.  I've tested this with some rough and ready sample code. I think this class should be delegated to rather than the `Connection` class when data is being uploaded...

MartinPaulo:
I've done some further investigation. `SwiftService` is a multithreaded class, and it uses a connection factory method to make sure that each thread can create connections on demand. Rather than trying to copy settings across to this factory method from the cloudbridge openstack provider, it would be better to override this factory method so that the `_connect_swift` factory was called instead to mint connections on demand.

This can be done by simply overriding the `swiftclient.service` module method that mints the connections in the cloudbridge `OpenStackBucketObject` as follows:

```python
 swiftclient.service.get_conn = self._provider._connect_swift
```

It would also be good if the `_connect_swift` method could accept and use the options that  `SwiftService` class passes to its connection factory.

MartinPaulo:
The `SwiftService` constructor builds its options from a grab bag of global, local and passed in options: and that it the filters out the options not needed for a connection by the way in which it calls its connection factory method. It does something along the lines of the following:

```python
class FakeConnection:
    def __init__(self, var1=None, var2=None, var3=None):
        self.var1 = var1
        self.var2 = var2
        self.var3 = var3

def get_fake_conn(f_options):
    return FakeConnection(
        f_options['var1'],
        f_options['var2'],
    )

grab_bag_of_options = {
    'var1': '1',
    'A': 'A',
    'var2': '2',
    'B': 'B',
}

connection = get_fake_conn(grab_bag_of_options)
```

So if the `_connect_swift` module is to accept options from the `SwiftService` instance it will also need to filter out options not used by the `Connection` constructor.

MartinPaulo:
[...]

PR #36, opened 2017-04-27
Upload delete of files larger than 5 Gig to Swift #35

MartinPaulo:
As coded, the low level Swift Connection is used to upload and delete
files. However, this class does not handle the upload and deletion of
files over 5 Gig in size that need to be segmented.

This commit replaces the use of the Swift Connection class with the
SwiftService class in these two cases. Files that are larger than
5 Gig will now be broken into segments on the 5 Gig boundary, and the
delete call will be able to delete the segmented files.

### DIFF
--- a/test/test_compute_service.py
+++ b/test/test_compute_service.py
@@ -384,18 +384,25 @@ class CloudComputeServiceTestCase(ProviderTestBase):
                 else:
                     external_net = net
                 router.attach_network(external_net.id)
-                # check whether adding an elastic ip works
-                fip = self.provider.network.create_floating_ip()
-                with helpers.cleanup_action(lambda: fip.delete()):
-                    test_inst.add_floating_ip(fip.public_ip)
-                    test_inst.refresh()
-                    self.assertIn(fip.public_ip, test_inst.public_ips)
+                router.add_route(subnet.id)
 
-                    if isinstance(self.provider, TestMockHelperMixin):
-                        # TODO: Moto bug does not refresh removed public ip
-                        return
+                def cleanup_router():
+                    router.remove_route(subnet.id)
+                    router.detach_network()
 
-                    # check whether removing an elastic ip works
-                    test_inst.remove_floating_ip(fip.public_ip)
-                    test_inst.refresh()
-                    self.assertNotIn(fip.public_ip, test_inst.public_ips)
+                with helpers.cleanup_action(lambda: cleanup_router()):
+                    # check whether adding an elastic ip works
+                    fip = self.provider.network.create_floating_ip()
+                    with helpers.cleanup_action(lambda: fip.delete()):
+                        test_inst.add_floating_ip(fip.public_ip)
+                        test_inst.refresh()
+                        self.assertIn(fip.public_ip, test_inst.public_ips)
+
+                        if isinstance(self.provider, TestMockHelperMixin):
+                            # TODO: Moto bug does not refresh removed public ip
+                            return
+
+                        # check whether removing an elastic ip works
+                        test_inst.remove_floating_ip(fip.public_ip)
+                        test_inst.refresh()
+                        self.assertNotIn(fip.public_ip, test_inst.public_ips)


MartinPaulo:
I see that Travis is not happy :(  I'll look into this tomorrow...

cloudbridge/cloud/providers/openstack/provider.py
```diff
@@ -236,19 +238,48 @@ def _connect_cinder(self):
...
+            result['os_options'] = None
+            passed_in_options = set(result.keys())
+            method_signature = inspect.signature(method_to_match)
+            parameters = set(method_signature.parameters.keys())
+            difference = passed_in_options - parameters
+            for name in difference:
+                del result[name]
+        return result
+
+    def connect_swift(self, options=None):
```
afgane: Yes, this is now being called from the outside class but I'm a bit hesitant to make this a public method. It adds asymmetry between the providers and may introduce a wrong impression on how methods should be called. So I'm thinking we may be better off leaving this private regardless of the outside call. @nuwang?

afgane:
This all looks very good. Thanks for taking a stab at it!

MartinPaulo:
I've made the requested changes. The tox flake8 commands now run flawlessly on my computer: however, the unit tests fail under tox because requests 2.12.5 is installed: and it clashes with the Babel!=2.4.0,>=2.3.4 requirement. If you do a clean pull and run of tox you get this as well. So there's another issue with conflicting dependencies that is causing this failure. If I run the tests manually, they pass just fine... I'll try to dig into the dependency graph to see if I can find out where the issue lies...

MartinPaulo:
Oh: and one other note about this fix: on our installation of Swift, the segments are put into their own segment bucket. The delete of a large file will not delete the associated segment bucket, even if there are no other large files with segments stored in it...

MartinPaulo:
I've dug into the failure and as a result of it:

* opened https://github.com/gvlproject/cloudbridge/issues/37
* found some more flaws in my code, hence the changes to the pull request.

As it stands this pull request won't pass the tests until the #37 is result :(

MartinPaulo:
I think I've totally managed to mess this pull request up by trying to be too clever with git. I'm going to close it and re-do it.

test/test_object_store_service.py
```diff
@@ -207,3 +209,33 @@ def test_upload_download_bucket_content_from_file(self):
...
+            with helpers.cleanup_action(lambda: test_bucket.delete()):
+                test_obj = test_bucket.create_object(file_name)
+                with helpers.cleanup_action(lambda: test_obj.delete()):
+                    file_uploaded = test_obj.upload_from_file(six_gig_file)
+                    self.assertTrue(file_uploaded, "Could not upload object?")
+                    with helpers.cleanup_action(
+                            lambda: os.remove(download_file)):
+                        with open(download_file, 'wb') as f:
+                            test_obj.save_content(f)
+                            self.assertTrue(
```
nuwang: @MartinPaulo I was wondering about the performance of this test. Does it take a long time? On the one hand, it would be great to test this. On the other, the test may take too long to complete. This may be an issue in particular with Travis, which has a build time-out of 45 minutes. What do you think? Can we perhaps makes the test optional or something?

MartinPaulo: I added the @skip annotation specifically to stop it being run, as it does indeed take a long time. Would you prefer something more Travis specific, such as: `@skipIf("TRAVIS" in os.environ and os.environ["TRAVIS"] == "true", "Skipping this test on Travis CI.")`?

nuwang: I missed that. I think we may as well leave it as is - but perhaps in future we can introduce a fast test and a rigorous test? Fast test for quick feedback during development and rigorous test for everything else?

MartinPaulo: No problem: I wasn't too fussed about a fast test, as this one is really just a restatement of the test before, but now with a very large object, as opposed to a small one.

Issue #37, opened 2017-05-01
Version Conflict when trying to run tox tests.


MartinPaulo:
If I clone the repository, setup the needed environment and then run tox, it fails under both py27 and py36 (I haven't tried pypy), with the following error message:

```bash
running test
Searching for requests!=2.12.2,!=2.13.0,>=2.10.0
# removed lines
Best match: requests 2.12.5
# removed lines
pkg_resources.VersionConflict: (Babel 2.4.0 (/home/travis/build/gvlproject/cloudbridge/.tox/py27-aws/lib/python2.7/site-packages), Requirement.parse('Babel!=2.4.0,>=2.3.4'))
```

Working through the dependencies in `setup.py` it would appear that indeed the version of requests that best fits them is 2.12.5. Which is why it is then downloaded and installed. However, that version of requests in turn explicitly states that it requires Babel==2.3.4 and not the version Babel==2.4.0 that has already been installed. Instant end of run :(

MartinPaulo:
My simple fix for this is to change `setup.py` to add the two required libraries:

```python
openstack_reqs = ['requests==2.12.5', 'Babel==2.3.4',
                  'python-novaclient==7.0.0',
                  'python-glanceclient>=2.5.0,<=2.6.0',
                  'python-cinderclient>=1.9.0,<=2.0.1',
                  'python-swiftclient>=3.2.0,<=3.3.0',
                  'python-neutronclient>=6.0.0,<=6.1.0',
                  'python-keystoneclient>=3.8.0,<=3.10.0']
```

Running the tests with this change in place gets them running. I'm not sure this is the best fix, though.

MartinPaulo:
I've dug into this a bit more - given my Java background this has been good learning. Sadly I can't find a better way, other than setting ranges on the two libraries (e.g.: `'requests<2.13.0', 'Babel>=2.3.4,<2.4.0'`)

nuwang:
@MartinPaulo Would you mind issuing a pull request for this?

MartinPaulo:
@nuwang No problem: will do

nuwang:
@MartinPaulo Thanks for fixing this. Closing issue.

PR #38, opened 2017-05-09
Version Conflict when trying to run tox tests #37

MartinPaulo:
If you clone the repository, setup the needed environment and then run
tox, it fails with a version conflict error.

This happens because the dependencies in setup.py are best resolved by
dropping down to and installing requests 2.12.5. However, that version
of requests in turn explicitly states that it requires Babel==2.3.4 and
not the version Babel==2.4.0 that has already been installed.

Instant end of run :(

This commit makes the version of Babel explicit

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,9 @@ with open(os.path.join('cloudbridge', '__init__.py')) as f:
             break
 
 base_reqs = ['bunch>=1.0.1', 'six>=1.10.0', 'retrying>=1.3.3']
-openstack_reqs = ['python-novaclient==7.0.0',
+openstack_reqs = ['requests<2.13.0',
+                  'Babel>=2.3.4,<2.4.0',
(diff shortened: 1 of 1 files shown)


MartinPaulo:
I've changed just one file: setup.py. When I run tox under python 3.5.1 in a virtualenv on my development machine against the nectar cloud, I get 2 failures, and 4 errors: all due to the tests not expecting the underlying nectar infrastructure ("Bad router request: Incompatible network.", "publicURL endpoint for compute service in Tasmania region not found" and "Newly created floating IP address should not be in use."). These are totally different failures to the one the build is reporting. I'm guessing that the test environment here has some other sort of issue?

nuwang:
@MartinPaulo Thanks for fixing this. I think it's ok if the tests don't pass agains NeCTAR because they are related to known networking issues. So I'll merge and close.

Issue #39, opened 2017-05-09
The tests expect block storage to be available in all zones


MartinPaulo:
As written, the tests for the block store service expect block storage to be available in the same default zone as the instance created. On our nectar cloud this is not necessarily true: block storage for a given project may only be available in one of the zones. Hence the tests attempt to create block storage in the same default zone as the instance will most often fail.

afgane:
This looks like it was fixed in #40 so closing the issue.

PR #40, opened 2017-05-09
The tests expect block storage to be available in all zones #39

MartinPaulo:
As written, the tests for the block store service expect block storage
to be available in the same default zone as the instance created.

On our nectar cloud this is not necessarily true: block storage for a
given project may only be available in one of the zones. Hence the tests
attempt to create block storage in the same default zone as the instance
will most often fail.

This commit allows the zone a test instance is to be run in to be set:
and changes two of the block store tests to use the
helpers.get_provider_test_data(self.provider, 'placement') setting. In
this way the instance can be made to run in the same zone that block
storage is available in.

### DIFF
--- a/test/test_block_store_service.py
+++ b/test/test_block_store_service.py
@@ -107,7 +107,9 @@ class CloudBlockStoreServiceTestCase(ProviderTestBase):
             net, subnet = helpers.create_test_network(
                 self.provider, instance_name)
             test_instance = helpers.get_test_instance(
-                self.provider, instance_name, subnet=subnet)
+                self.provider, instance_name, subnet=subnet,
+                zone=helpers.get_provider_test_data(self.provider,
+                                                    'placement'))
             name = "CBUnitTestAttachVol-{0}".format(uuid.uuid4())
             test_vol = self.provider.block_store.volumes.create(
                 name, 1, test_instance.zone_id)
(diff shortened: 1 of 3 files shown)


MartinPaulo:
As expected, the build failed due to the Version Conflict when trying to run tox tests issue ( #37 )

test/test_block_store_service.py
```diff
@@ -107,7 +107,9 @@ def test_attach_detach_volume(self):
             net, subnet = helpers.create_test_network(
                 self.provider, instance_name)
             test_instance = helpers.get_test_instance(
-                self.provider, instance_name, subnet=subnet)
+                self.provider, instance_name, subnet=subnet,
+                zone=helpers.get_provider_test_data(self.provider,
```
nuwang: @MartinPaulo What if we remove the zone parameter from the signature and let the get_test_instance method internally default to using the zone from provider_test_data? This is already being done this way for machine type and image in create_test_instance, so it may make it simpler to use and consistent?

MartinPaulo: No problem.

MartinPaulo:
Ok: I've added in the fix for #37 and moved the zone parameter from the signature and let the get_test_instance method internally default to using the zone from provider_test_data, as suggested. At least the tests now run. Again, as per #37, there are only 3 failures on my machine when run against the nectar cloud, as opposed to the large number of failures experienced in this build :(

nuwang:
@MartinPaulo Looks great! Thanks for fixing all of this.

PR #41, opened 2017-05-10
Upload delete of files larger than 5 Gig to Swift #35

MartinPaulo:
As coded, the low level Swift Connection is used to upload and delete
files. However, this class does not handle the upload and deletion of
files over 5 Gig in size that need to be segmented.

This commit replaces the use of the Swift Connection class with the
SwiftService class in these two cases. Files that are larger than
5 Gig will now be broken into segments on the 5 Gig boundary, and the
delete call will be able to delete the segmented files.

### DIFF
--- a/cloudbridge/cloud/providers/openstack/provider.py
+++ b/cloudbridge/cloud/providers/openstack/provider.py
@@ -244,10 +244,12 @@ class OpenStackCloudProvider(BaseCloudProvider):
         Returns a **copy** of the source options with all keys that are not in
         the ``method_to_match`` parameter list removed.
 
-        .. note:: If ``options`` has the ``os_options`` key set it will have
-            any values the value dictionary removed in they copy if they are
-            not set, as they have higher precedence that our settings and thus
-            effectively mask our settings.
+        .. note:: If ``options`` has the ``os_options`` key it will have
+            both the key and its value removed. This is because any entries
+            in this dictionary value will override our settings. This
+            situation is only going to happen when the `_connect_swift`
+            method is called by the SwiftService to manufacture new
+            connections.
 
         .. seealso::
             https://docs.openstack.org/developer/python-swiftclient/swiftclient.html#module-swiftclient.client
@@ -271,9 +273,7 @@ class OpenStackCloudProvider(BaseCloudProvider):
             result = {key: val for key, val in options.items() if
                       key in parameters}
             # Don't allow the options to override our authentication
-            if 'os_options' in result:
-                result['os_options'] = {key: val for key, val in
-                                        result['os_options'].items() if val}
+            result.pop('os_options', None)
         return result
 
     def _connect_swift(self, options=None):
--- a/cloudbridge/cloud/providers/openstack/provider.py
+++ b/cloudbridge/cloud/providers/openstack/provider.py
@@ -241,8 +241,16 @@ class OpenStackCloudProvider(BaseCloudProvider):
     @staticmethod
     def _clean_options(options, method_to_match):
         """
-        Returns a copy of the source options with all keys that are not in the
-        ``method_to_match`` parameter list removed.
+        Returns a **copy** of the source options with all keys that are not in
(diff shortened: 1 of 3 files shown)


MartinPaulo:
As with my other commits, when I run tox under python 3.5.1 in a virtualenv on my development machine against the nectar cloud, I get 2 failures, and 4 errors: all due to the tests not expecting the underlying nectar infrastructure ("Bad router request: Incompatible network.", "publicURL endpoint for compute service in Tasmania region not found" and "Newly created floating IP address should not be in use."). These are totally different failures to the one the build is reporting. I'm guessing that the test environment here has some other sort of issue?

afgane:
The tests are failing because this is coming from a fork, evident from the env:
- PR: https://travis-ci.org/gvlproject/cloudbridge/jobs/230596467#L160
- Direct commit: https://travis-ci.org/gvlproject/cloudbridge/jobs/232989289#L149

There's an open issue for this on Travis: https://github.com/travis-ci/travis-ci/issues/1946

afgane:
Ohh, and just wondering if you'd have any interest in adding support for uploading large files to the AWS side of the implementation (ie https://github.com/gvlproject/cloudbridge/blob/master/cloudbridge/cloud/providers/aws/resources.py#L845)?

MartinPaulo:
Re the failing tests: it seems like Travis are moving very slowly to any form of a fix :(  It kind of negates the benefit of Travis in this environment IMHO.

Re the AWS side of the implementation: I don't have time at the moment. Perhaps in another month or so :(

cloudbridge/cloud/providers/openstack/provider.py
```diff
@@ -236,19 +238,58 @@ def _connect_cinder(self):
...
+        result = dict(options or {})
+        if len(result):
+            # Don't allow the options to override our authentication
+            result['os_options'] = None
+            passed_in_options = set(result.keys())
+            try:
+                method_signature = inspect.signature(method_to_match)
+                parameters = set(method_signature.parameters.keys())
+            except AttributeError:
+                parameters = set(inspect.getargspec(method_to_match)[0])
```
nuwang: @MartinPaulo Just a small suggestion for clarity: parameters = set(inspect.getargspec(method_to_match).args) instead of [0].

cloudbridge/cloud/providers/openstack/provider.py
```diff
@@ -236,19 +238,58 @@ def _connect_cinder(self):
...
+            difference = passed_in_options - parameters
+            for name in difference:
+                del result[name]
+        return result
```
nuwang: @MartinPaulo A dictionary comprehension based implementation seems to shorten the code slightly. Do you foresee any issues?
```
def _clean_options(options, method_to_match):
    if options:
        try:
              method_signature = inspect.signature(method_to_match)
              parameters = set(method_signature.parameters.keys())
        except AttributeError:
              parameters = set(inspect.getargspec(method_to_match).args)
        result = { key : val for key, val in options.items() if key not in parameters }
        # Don't allow the options to override our authentication
        result['os_options'] = None
        return result
     else:
         return options or {}
```

cloudbridge/cloud/providers/openstack/provider.py
MartinPaulo: Yes; a much better way of doing it...

cloudbridge/cloud/providers/openstack/provider.py
MartinPaulo: I want to make sure that one who maintains this code will not return the original options: as if they are mutated the changes might affect other operations elsewhere. Outside of that (and I think it should be `if key in parameters`) this does make the code more readable. I'll make the change and resubmit...

cloudbridge/cloud/providers/openstack/provider.py
MartinPaulo: I'll make the change and resubmit..

Issue #42, opened 2017-05-10
OpenStackFloatingIP in_use method doesn't necessarily return the correct result


MartinPaulo:
When run against the nectar cloud the test_crud_network_service method fails as follows.

```bash
======================================================================
FAIL: test_crud_network_service (test.test_network_service.CloudNetworkServiceTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/mpaulo/PycharmProjects/cloudbridge/test/helpers.py", line 72, in wrapper
    func(self, *args, **kwargs)
  File "/Users/mpaulo/PycharmProjects/cloudbridge/test/test_network_service.py", line 85, in test_crud_network_service
    "Newly created floating IP address should not be in use.")
AssertionError: Newly created floating IP address should not be in use.
```
When digging into the code it is found that the OpenStackFloatingIP in_use method tests the floating IP status, and if it is found to be 'ACTIVE' returns True. However, on our nectar cloud a newly created floating IP is set to 'ACTIVE' by default - even though it is not in use. A quick trawl through the neutron code seems to show that the status of a newly created floating IP depends on the underlying hardware. Hence ACTIVE is not a good indicator as to if the IP is in use or not.

MartinPaulo:
I've been trying to work out what indicates if a neutron floating ip is in use or not. The guys here say "port_id is set". Looking at the shade code  [shade/shade/_normalize.py](https://github.com/openstack-infra/shade/blob/fcf76a9ca981ff4731e3c1f7b58729a04ed13f37/shade/_normalize.py) the `_normalize_floating_ip` method has the following line `attached = bool(port_id)`, so it looks like the guys are correct...

afgane:
Fixed in #43.

PR #43, opened 2017-05-10
OpenStackFloatingIP in_use method doesn't return the correct result #42

MartinPaulo:
The floating IP status was tests, and if found to be 'ACTIVE' True was
returned. However, on our nectar cloud a newly created floating IP is
set to 'ACTIVE' by default - even though it is not in use.

This commit uses port id as indicator of being in use: if set, then
the floating ip is attached, if not, it is not.

### DIFF
--- a/cloudbridge/cloud/providers/openstack/resources.py
+++ b/cloudbridge/cloud/providers/openstack/resources.py
@@ -788,7 +788,7 @@ class OpenStackFloatingIP(BaseFloatingIP):
         return self._ip.get('fixed_ip_address', None)
 
     def in_use(self):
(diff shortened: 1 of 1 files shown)


MartinPaulo:
As with my other commits, when I run tox under python 3.5.1 in a virtualenv on my development machine against the nectar cloud, I get a few failures/errors: all due to the tests not expecting the underlying nectar infrastructure ("Bad router request: Incompatible network." and "publicURL endpoint for compute service in Tasmania region not found". These are totally different failures to the ones the build is reporting. I'm guessing that the test environment here has some other sort of issue?

afgane:
I've tested this locally for now and it's working as you described. Thanks for following through with it.

PR #44, opened 2017-05-12
Add GCEBlockStoreService, GCEVolumeService, GCESnapshotService

baizhang:

This pull request adds GCEBlockStoreService, GCEVolumeService, GCESnapshotService and related classes.

Thank you!

### DIFF
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -17,6 +17,7 @@ import googleapiclient.http
 from oauth2client.client import GoogleCredentials
 from oauth2client.service_account import ServiceAccountCredentials
 
+from .services import GCEBlockStoreService
 from .services import GCEComputeService
 from .services import GCENetworkService
 from .services import GCESecurityService
@@ -35,7 +36,7 @@ class GCPResourceUrl(object):
 
 
 class GCPResources(object):
-    
+
     def __init__(self, connection):
         self._connection = connection
 
@@ -82,6 +83,8 @@ class GCPResources(object):
 
         # We will not mutate self._desc; it's OK to use items() in Python 2.x.
         for resource, resource_desc in desc['resources'].items():
+            if 'methods' not in resource_desc:
+                continue
             methods = resource_desc['methods']
             if 'get' not in methods:
                 continue
@@ -161,6 +164,7 @@ class GCECloudProvider(BaseCloudProvider):
         self._compute = GCEComputeService(self)
         self._security = GCESecurityService(self)
         self._network = GCENetworkService(self)
+        self._block_store = GCEBlockStoreService(self)
 
         self._compute_resources = GCPResources(self.gce_compute)
         self._storage_resources = GCPResources(self.gcp_storage)
@@ -179,8 +183,7 @@ class GCECloudProvider(BaseCloudProvider):
 
     @property
     def block_store(self):
-        raise NotImplementedError(
-            "GCECloudProvider does not implement this service")
+        return self._block_store
 
     @property
     def object_store(self):
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -1311,7 +1311,6 @@ class GCEVolume(BaseVolume):
 
     @property
     def description(self):
-        self.refresh()
         labels = self._volume.get('labels')
         if not labels or 'description' not in labels:
             return ''
@@ -1319,18 +1318,21 @@ class GCEVolume(BaseVolume):
 
     @description.setter
     def description(self, value):
-        self.refresh()
         request_body = {
             'labels': {'description': value.replace(' ', '_').lower(),},
             'labelFingerprint': self._volume.get('labelFingerprint'),
         }
-        response = (self._provider.gce_compute
-                    .disks()
-                    .setLabels(
-                        project=self._provider.project_name,
-                        zone=self._provider.default_zone,
-                        resource=self.name,
-                        body=request_body).execute())
+        try:
+            response = (self._provider.gce_compute
+                        .disks()
+                        .setLabels(
+                            project=self._provider.project_name,
+                            zone=self._provider.default_zone,
+                            resource=self.name,
+                            body=request_body).execute())
+        except Exception as e:
+            cb.log.warning('Exception while setting volume description: %s', e)
+        self.refresh()
 
     @property
     def size(self):
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -1311,11 +1311,26 @@ class GCEVolume(BaseVolume):
 
     @property
     def description(self):
-        return self._volume.get('description')
+        self.refresh()
+        labels = self._volume.get('labels')
+        if not labels or 'description' not in labels:
+            return ''
+        return labels.get('description', '')
 
     @description.setter
     def description(self, value):
-        raise NotImplementedError('Not supported by this provider.')
+        self.refresh()
+        request_body = {
+            'labels': {'description': value.replace(' ', '_').lower(),},
+            'labelFingerprint': self._volume.get('labelFingerprint'),
+        }
+        response = (self._provider.gce_compute
+                    .disks()
+                    .setLabels(
+                        project=self._provider.project_name,
+                        zone=self._provider.default_zone,
+                        resource=self.name,
+                        body=request_body).execute())
 
     @property
     def size(self):
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -1,6 +1,7 @@
 """
 DataTypes used by this provider
 """
+from cloudbridge.cloud.base.resources import BaseAttachmentInfo
 from cloudbridge.cloud.base.resources import BaseFloatingIP
 from cloudbridge.cloud.base.resources import BaseInstance
 from cloudbridge.cloud.base.resources import BaseInstanceType
@@ -11,8 +12,12 @@ from cloudbridge.cloud.base.resources import BasePlacementZone
 from cloudbridge.cloud.base.resources import BaseRegion
 from cloudbridge.cloud.base.resources import BaseSecurityGroup
 from cloudbridge.cloud.base.resources import BaseSecurityGroupRule
+from cloudbridge.cloud.base.resources import BaseSnapshot
+from cloudbridge.cloud.base.resources import BaseVolume
 from cloudbridge.cloud.interfaces.resources import InstanceState
 from cloudbridge.cloud.interfaces.resources import MachineImageState
+from cloudbridge.cloud.interfaces.resources import SnapshotState
+from cloudbridge.cloud.interfaces.resources import VolumeState
 
 import cloudbridge as cb
 
@@ -80,7 +85,7 @@ class GCEInstanceType(BaseInstanceType):
 
     @property
     def id(self):
-        return str(self._inst_dict.get('id'))
+        return self._inst_dict.get('selfLink')
 
     @property
     def name(self):
@@ -193,7 +198,7 @@ class GCERegion(BaseRegion):
 class GCEFirewallsDelegate(object):
     DEFAULT_NETWORK = 'default'
     _NETWORK_URL_PREFIX = 'global/networks/'
-  
+
     def __init__(self, provider):
         self._provider = provider
         self._list_response = None
@@ -222,7 +227,7 @@ class GCEFirewallsDelegate(object):
             if network_name is not None:
                 out.add((firewall['targetTags'][0], network_name))
         return out
-            
+
     def network_name(self, firewall):
         """
         Extract the network name of a firewall.
@@ -637,7 +642,7 @@ class GCEMachineImage(BaseMachineImage):
         :rtype: ``str``
         :return: ID for this instance as returned by the cloud middleware.
         """
-        return self._gce_image['name']
+        return self._gce_image.get('selfLink')
 
     @property
     def name(self):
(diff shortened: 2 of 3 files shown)


cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1063,3 +1068,253 @@ def delete(self):
...
+
+    @property
+    def name(self):
+        """
+        Get the volume name.
+        """
+        return self._volume.get('name')
+
+    @name.setter
+    # pylint:disable=arguments-differ
```
mbookman: Why is this pylint directive needed? Why isn't it needed for the description setter?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1063,3 +1068,253 @@ def delete(self):
...
+
+    @property
+    def source(self):
+        return self._volume.get('sourceSnapshot')
+
+    @property
+    def attachments(self):
+        if 'users' in self._volume and len(self._volume) > 0:
+            return BaseAttachmentInfo(self,
+                                      self._volume.get('users')[0],
```
mbookman: Why explicitly get argument 0 here?
PD does support multiple instances attaching a disk, so long as they are all read-only attachments.
If the assumption here is that there is a single instance that has attached the disk read-write, then please add a comment to that effect.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1063,3 +1068,253 @@ def delete(self):
...
+                                      None)
+        else:
+            return None
+
+    def attach(self, instance, device):
+        """
+        Attach this volume to an instance.
+
+        Parameter `device` is ignored here. The user need to mount the disk
```
mbookman: s/need/needs

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1063,3 +1068,253 @@ def delete(self):
...
+        so that the operating system can use the available storage space.
+        https://cloud.google.com/compute/docs/disks/add-persistent-disk
+        """
+        attach_disk_body = {
+            "source": self.id
+        }
+        instance_name = instance.name if isinstance(
+            instance,
```
mbookman: `instance` could be either a GCEInstance or a string containing the name?
That should be clearly documented in the args section of the function.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1063,3 +1068,253 @@ def delete(self):
...
+                            project=self._provider.project_name,
+                            zone=self._provider.default_zone,
+                            instance=instance_name,
+                            body=attach_disk_body).execute())
+
+    def detach(self, force=False):
+        """
+        Detach this volume from an instance.
+        """
+        if not self.attachments:
```
mbookman: Please add some commenting here about this first block where you are checking multiple ways whether the volume is actually attached to an instance. It isn't immediately obvious why all the checks are needed.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -661,3 +667,196 @@ def routers(self):
...
+                    for vol in response['items']]
+        return ServerPagedResultList(len(gce_vols) > max_result,
+                                     response.get('nextPageToken'),
+                                     False, data=gce_vols)
+
+    def list(self, limit=None, marker=None):
+        """
+        List all volumes.
+        """
+        # For GCE API, Acceptable values are 0 to 500, inclusive.
```
mbookman: It is not completely obvious without going to the API documentation that this comment (and maxResults) are about response paging and not about completely limiting the total number of records returned. In fact it is unclear what the "limit" parameter means- per-page results or total results. Please add comments to clarify.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1063,3 +1068,253 @@ def delete(self):
...
+        """
+        return self._volume.get('name')
+
+    @name.setter
+    # pylint:disable=arguments-differ
+    def name(self, value):
+        """
+        Set the volume name.
+        """
+        raise NotImplementedError('Not supported by this provider.')
```
nuwang: @baizhang @mbookman I took a look at the API and it looks like the name in GCE actually corresponds to the ID in CloudBridge, that is to say, the name is the primary key/ID that's used to fetch a disk, and therefore, cannot be changed after creation. Therefore, can we use a tag or something to simulate a name and description for GCE disks?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1063,3 +1068,253 @@ def delete(self):
...
+    @property
+    def attachments(self):
+        if 'users' in self._volume and len(self._volume) > 0:
+            return BaseAttachmentInfo(self,
+                                      self._volume.get('users')[0],
+                                      None)
+        else:
+            return None
+
+    def attach(self, instance, device):
```
nuwang: @baizhang @mbookman The device here means the blk device id, and not the mount point name. I think that the attach method will need to use or return a device id as otherwise, there would be no way to subsequently mount a disk without knowing the device id. This post suggests that it's the device_name parameter: https://stackoverflow.com/questions/33210664/programmatically-createmount-disk-from-within-google-compute-vm

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1063,3 +1068,253 @@ def delete(self):
...
+         """
+        return self._snapshot.get('name')
+
+    @name.setter
+    # pylint:disable=arguments-differ
+    def name(self, value):
+        """
+        Set the snapshot name.
+        """
+        raise NotImplementedError('Not supported by this provider.')
```
nuwang: @baizhang @mbookman Same as above, perhaps we can simulate a name and description using tags?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -661,3 +667,196 @@ def routers(self):
...
+    def create(self, name, size, zone, snapshot=None, description=None):
+        """
+        Creates a new volume.
+
+        Argument `name` must be 1-63 characters long, and comply with RFC1035.
+        Specifically, the name must be 1-63 characters long and match the
+        regular expression [a-z]([-a-z0-9]*[a-z0-9])? which means the first
+        character must be a lowercase letter, and all following characters must
+        be a dash, lowercase letter, or digit, except the last character, which
+        cannot be a dash.
```
nuwang: @baizhang @mbookman Thanks for documenting this. It once again suggests to me that the name parameter in GCE cannot be mapped to the name property in cloudbridge, since name in cloudbridge is meant for an end user, and is not meant to be an unchangeable ID. Enforcing that users enter a name starting with a lowercase character for volume names across cloudbridge would be problematic. Therefore, we could either try to simulate a name (say using tags), or apply a fixed transformation to a user provided name so that it conforms to GCE requirements. What do you think?

cloudbridge/cloud/providers/gce/resources.py
mbookman: Hi Nuwang,

Seems like using the [setLabels](https://cloud.google.com/compute/docs/reference/beta/disks/setLabels) beta API could achieve what you want here so long as the name/description functionality for the volumes are really just labels. In other words there is no lookup by name or similar functionality.

mbookman:
@nuwang - We aren't really sure how the Volume name is used in the API or used by Galaxy - is it okay for it to be just a label? Are lookups by "name" needed?

nuwang:
@mbookman At present, the volume label is mainly used as a display value - cloudbridge/cloudman will internally use the volume ID as primary key, but volume name will be displayed for the users' benefit. Currently, there's no scenario in which lookups by name are needed. However, I would imagine that a search by name may be needed at some point. Do labels support searching? I guess we could always simulate a search within cloudbridge itself - i.e. retrieve all volumes, get their name label, and then filter the ones matching the search criteria  - would that be terribly slow?

Enis and I had a chat about this yesterday and we were going to suggest that perhaps the thing to do for GCE is to autogenerate a UUID for the Volume ID, and then use a label for the volume name. Would that work?

cloudbridge/cloud/providers/gce/services.py
mbookman: I'm not sure there are any good options for GCE if you want to allow for free-form names and descriptions on Volumes. The Name is going to be restricted by what Bai has documented here:

The name must be 1-63 characters long, and comply with RFC1035.

However labels have the same [restrictions](https://cloud.google.com/compute/docs/labeling-resources#restrictions):

- Keys and values cannot be longer than 63 characters each.
- Keys and values can only contain lowercase letters, numeric characters, underscores, and dashes. International characters are allowed.

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -82,6 +83,8 @@ def __init__(self, connection):
 
         # We will not mutate self._desc; it's OK to use items() in Python 2.x.
         for resource, resource_desc in desc['resources'].items():
+            if 'methods' not in resource_desc:
```
baizhang: Done.

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/resources.py

cloudbridge/cloud/providers/gce/services.py

cloudbridge/cloud/providers/gce/resources.py
baizhang: It seems that even if we simulate name and description using labels, similar restrictions will apply. Please see mbookman's comment below.

baizhang:
Hi @mbookman and @nuwang ,

Thank you for your review comments. Please take another look. Thanks!

nuwang:
@baizhang @mbookman @afgane Thanks for the changes and suggestions. Looks good. I was just thinking though, that if we use the name instead of label, we will not be able to rename the volume later I'm assuming? That's probably ok, we can just document that as not being an option in GCE.

All in all, based on what you said Matt, I was thinking that perhaps we should consider enforcing RFC1035 for all names in cloudbridge and document it at an interface level. The only thing that bugs me a little is the lack of spaces for a descriptive name. Perhaps it can be "RFC1035+spaces allowed" and GCE can replace spaces with underscores where necessary?

mbookman:
@nuwang - your proposal seems reasonable to me:

- Use the disk name field for name (with the specified character and inability to rename restrictions)
- Use labels for the disk description field. GCE implementation will convert spaces to underscores

baizhang:
Hi @nuwang @mbookman 

Commit 7df11e4 adds GCEVolume description using labels. PTAL.

Thanks!

mbookman:
Thanks @baizhang. Commit 7df11e4 looks good to me so long as the caller is going to be enforcing restrictions on the characters and length of the description. Put another way - what happens if the GCEVolume description has invalid label characters - does the error message raised help the end user or do they get something cryptic?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1311,11 +1311,26 @@ def name(self, value):
 
     @property
     def description(self):
-        return self._volume.get('description')
+        self.refresh()
```
afgane: We had a bit of a discussion on this particular `refresh` and concluded that unless there is a technical necessity to keep it, it'd be better to remove it as a way to keep the response time and the number of requests down. We're going to try and standardize on that across the library and make it explicit that the user code should call the object's `refresh` method rather than making a decision when the library ought to do it.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1311,11 +1311,26 @@ def name(self, value):
...
+        labels = self._volume.get('labels')
+        if not labels or 'description' not in labels:
+            return ''
+        return labels.get('description', '')
 
     @description.setter
     def description(self, value):
-        raise NotImplementedError('Not supported by this provider.')
+        self.refresh()
```
afgane: Same here as above. This actually has me wondering why is the refresh here to begin with since the value will be set (hence, overwritten) in the next couple of lines?

cloudbridge/cloud/providers/gce/resources.py

baizhang:
Hi @afgane @mbookman @nuwang 

I updated the this pull request. For @afgane 's comments, please see inline replies. For @mbookman 's comment, I log the error message to cb.log.warning when there are invalid label characters.

PTAL. Thanks!

cloudbridge/cloud/providers/gce/resources.py
baizhang: The reason is that to add or update labels, the setLabels method needs to provide with the latest fingerprint and new labels. Here is a relevant document on this topic.
https://cloud.google.com/compute/docs/labeling-resources

In commit e300d1b, I moved the refresh call to the end of the description setter, so that self._volume will hold the latest data of the volume resource (including the fingerprint) after a new label is set. I also removed the refresh call in the description getter, with the assumption that cloudbridge users only interact with GCE resources through cloudbridge.

In this way, two consecutive description setter calls will work naturally, and consecutive description setter and getter calls will work as well, without the need to call refresh in between.

nuwang:
@baizhang I'd be more inclined to log the message but re-raise the error, rather than failing silently. If it's ok, I'll merge this in and issue a new pull request to fix that.

nuwang:
@afgane @baizhang @mbookman Have included aforementioned fix in: https://github.com/gvlproject/cloudbridge/commit/34ae3bd7b307d3f3f2d5fb1af87a0241f65ede8e

PR #50, opened 2017-06-26
GCP routers, subnets, buckets, and objects

chiniforooshan:
@nuwang @afgane @baizhang @mbookman 

Please take a look.

### DIFF
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -1,6 +1,13 @@
 """
 DataTypes used by this provider
 """
+import hashlib
+import inspect
+import json
+import re
+import uuid
+
+import cloudbridge as cb
 from cloudbridge.cloud.base.resources import BaseAttachmentInfo
 from cloudbridge.cloud.base.resources import BaseFloatingIP
 from cloudbridge.cloud.base.resources import BaseInstance
@@ -10,28 +17,25 @@ from cloudbridge.cloud.base.resources import BaseMachineImage
 from cloudbridge.cloud.base.resources import BaseNetwork
 from cloudbridge.cloud.base.resources import BasePlacementZone
 from cloudbridge.cloud.base.resources import BaseRegion
+from cloudbridge.cloud.base.resources import BaseRouter
 from cloudbridge.cloud.base.resources import BaseSecurityGroup
 from cloudbridge.cloud.base.resources import BaseSecurityGroupRule
 from cloudbridge.cloud.base.resources import BaseSnapshot
+from cloudbridge.cloud.base.resources import BaseSubnet
 from cloudbridge.cloud.base.resources import BaseVolume
 from cloudbridge.cloud.interfaces.resources import InstanceState
 from cloudbridge.cloud.interfaces.resources import MachineImageState
+from cloudbridge.cloud.interfaces.resources import RouterState
 from cloudbridge.cloud.interfaces.resources import SnapshotState
 from cloudbridge.cloud.interfaces.resources import VolumeState
 
-import cloudbridge as cb
+import googleapiclient
 
 # Older versions of Python do not have a built-in set data-structure.
 try:
     set
 except NameError:
     from sets import Set as set
-
-import hashlib
-import inspect
-import json
-import re
-import uuid
 
 
 class GCEKeyPair(BaseKeyPair):
@@ -187,8 +191,11 @@ class GCERegion(BaseRegion):
         """
         Accesss information about placement zones within this region.
         """
-        zones_response = self._provider.gce_compute.zones().list(
-            project=self._provider.project_name).execute()
+        zones_response = (self._provider
+                              .gce_compute
+                              .zones()
+                              .list(project=self._provider.project_name)
+                              .execute())
         zones = [zone for zone in zones_response['items']
                  if zone['region'] == self._gce_region['selfLink']]
         return [GCEPlacementZone(self._provider, zone['name'], self.name)
@@ -285,11 +292,12 @@ class GCEFirewallsDelegate(object):
             firewall['sourceTags'] = [source_tag]
         project_name = self._provider.project_name
         try:
-            response = (self._provider.gce_compute
-                                      .firewalls()
-                                      .insert(project=project_name,
-                                              body=firewall)
-                                      .execute())
+            response = (self._provider
+                            .gce_compute
+                            .firewalls()
+                            .insert(project=project_name,
+                                    body=firewall)
+                            .execute())
             self._provider.wait_for_operation(response)
             # TODO: process the response and handle errors.
         except:
@@ -314,7 +322,8 @@ class GCEFirewallsDelegate(object):
             if not self._check_list_in_dict(firewall, 'sourceRanges',
                                             source_range):
                 continue
-            if not self._check_list_in_dict(firewall, 'sourceTags', source_tag):
+            if not self._check_list_in_dict(firewall, 'sourceTags',
+                                            source_tag):
                 continue
             return firewall['id']
         return None
@@ -328,7 +337,7 @@ class GCEFirewallsDelegate(object):
             if firewall['id'] != firewall_id:
                 continue
             if ('sourceRanges' in firewall and
-                len(firewall['sourceRanges']) == 1):
+                    len(firewall['sourceRanges']) == 1):
                 info['source_range'] = firewall['sourceRanges'][0]
             if 'sourceTags' in firewall and len(firewall['sourceTags']) == 1:
                 info['source_tag'] = firewall['sourceTags'][0]
@@ -337,7 +346,7 @@ class GCEFirewallsDelegate(object):
             if 'IPProtocol' in firewall['allowed'][0]:
                 info['ip_protocol'] = firewall['allowed'][0]['IPProtocol']
             if ('ports' in firewall['allowed'][0] and
-                len(firewall['allowed'][0]['ports']) == 1):
+                    len(firewall['allowed'][0]['ports']) == 1):
                 info['port'] = firewall['allowed'][0]['ports'][0]
             info['network_name'] = self.network_name(firewall)
             return info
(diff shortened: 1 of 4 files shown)


cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -666,9 +680,12 @@ def delete(self):
         """
         Delete this image
         """
-        request = self._provider.gce_compute.images().delete(
-            project=self._provider.project_name, image=self.name)
-        request.execute()
+        (self._provider
```
mbookman: Why is this call wrapped in parentheses?
I see a lot of similar such cases in this PR.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -654,28 +683,138 @@ def create_floating_ip(self, region=None):
...
+                            .gce_compute
+                            .routers()
+                            .insert(project=self.provider.project_name,
+                                    region=region,
+                                    body={'name': name,
+                                          'network': network_url})
+                            .execute())
+            if 'error' in response:
+                return None
+            self.provider.wait_for_opeartion(response, region=region)
```
mbookman: s/opeartion/operation

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -872,3 +1024,76 @@ def create(self, name, volume, description=None):
...
+        does not exist or if the user does not have permission to access the
+        bucket.
+        """
+        try:
+            response = (self.provider
+                            .gcp_storage
+                            .buckets()
+                            .get(bucket=bucket_id)
+                            .execute())
+            if 'error' in response:
```
mbookman: Should this check for 403 and 404 and otherwise raise an exception? log an error?

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: For breaking long lines, using extra parantheses around the expression is preferred to using backslashes, except for "with".

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Done.

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Done.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1548,11 +1684,163 @@ def create_volume(self, placement, size=None, volume_type=None, iops=None):
...
+        return self._obj['size']
+
+    @property
+    def last_modified(self):
+        return self._obj['updated']
+
+    def iter_content(self):
+        # TODO: It's not clear what does this method do. Does it return an
+        # iterator for metadata fields?
+        raise NotImplementedError('Not Implemented')
```
nuwang: @chiniforooshan This allows the actual content of the object to be iterated through as a byte stream, for example as shown here: https://github.com/gvlproject/cloudbridge/blob/6da5b0057385d7e4ba23245a5db9d8139513663b/cloudbridge/cloud/base/resources.py#L604

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -872,3 +1024,80 @@ def create(self, name, volume, description=None):
...
+            response = (self.provider
+                            .gcp_storage
+                            .buckets()
+                            .list(project=self.provider.project_name)
+                            .execute())
+            if 'error' in response or 'items' not in response:
+                return []
+            buckets = [GCSBucket(self.provider, bucket)
+                       for bucket in response['items']]
+            return ClientPagedResultList(self.provider, buckets,
```
nuwang: @chiniforooshan I think it would be much better to use a ServerPagedResultList so that filtering happens more efficiently. I believe the bucket.list() method supports maxResults and pageToken which would make this possible.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Thanks. This is implemented now.

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Good point. Thanks. I switched to ServerPagedResultList. While doing it looked to me that the current usage of ServerPagedResultList is not right in some places. Those are fixed, too.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -654,28 +708,148 @@ def create_floating_ip(self, region=None):
...
+        super(GCESubnetService, self).__init__(provider)
 
-    def create_router(self, name=None):
+    def get(self, subnet_id):
+        for subnet in self.list():
+            if subnet.id == subnet_id:
+                return subnet
+        return None
+
+    def list(self, network=None, region=None):
```
nuwang: @chiniforooshan @afgane The list method needs the marker and limit parameters as per the interface. The aws and openstack implementations also have the same bug here, because they don't accept limit and marker. We need an accompanying test which detects this. With limit and marker, we can also simplify the implementation below because there's no need to iterate through all the results.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -396,21 +411,33 @@ def _update_list_response(self):
...
+                                 .execute())
+            if self._list_response is None:
+                self._list_response = list_response
+                if 'items' not in self._list_response:
+                    self._list_response['items'] = []
+            else:
+                self._list_response['items'] += list_response.get('items', [])
+            if 'nextPageToken' not in list_response:
+                return
+            token = list_response['nextPageToken']
```
nuwang: @chiniforooshan I feel that there are several places where this overall pattern of logic is being repeated. I wonder whether it can be factored out into a helper like with the PageableObjectMixin: https://github.com/gvlproject/cloudbridge/blob/master/cloudbridge/cloud/base/resources.py#L200

So that teration can simply be something like:
```
for item in gce_helpers.iter_all(self._provider.gce_compute.firewalls(), project=blah, pageToken=token):
   do something
```

and iter_all could be something like:
```
def iter_all(gce_resource, **kwargs):
    do:
       response =   gce_resource.list(kwargs)
       if 'items' not in response:
             return
       for item in response['items']:
              yield item
        token = response.get('nextPageToken')
    while token
```

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -927,18 +961,24 @@ def _get_existing_target_instance(self):
...
+                                .execute())
+                if 'items' not in response:
+                    break
+                for target_instance in response['items']:
+                    url = self._provider.parse_url(target_instance['instance'])
+                    if url.parameters['instance'] == self.name:
+                        return target_instance
+                if 'nextPageToken' not in response:
+                    break
+                token = response['nextPageToken']
```
nuwang: @chiniforooshan Can the suggestion at line 431 be applied here?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -654,28 +708,148 @@ def create_floating_ip(self, region=None):
...
+                                .routers()
+                                .list(project=self.provider.project_name,
+                                      region=region,
+                                      pageToken=token)
+                                .execute())
+                for router in response.get('items', []):
+                    routers.append(GCERouter(self.provider, router))
+                if 'nextPageToken' not in response:
+                    break
+                token = response['nextPageToken']
```
nuwang: @chiniforooshan Another line 431 candidate?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -628,21 +676,27 @@ def create(self, name):
...
+                                .addresses()
+                                .list(project=self.provider.project_name,
+                                      region=region,
+                                      pageToken=token)
+                                .execute())
+                for ip in response.get('items', []):
                     ips.append(GCEFloatingIP(self.provider, ip))
+                if 'nextPageToken' not in response:
+                    break
+                token = response['nextPageToken']
```
nuwang: @chiniforooshan Another line 431 candidate?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -654,28 +708,148 @@ def create_floating_ip(self, region=None):
...
+                                .subnetworks()
+                                .list(project=self.provider.project_name,
+                                      region=region,
+                                      pageToken=token)
+                                .execute())
+                for subnet in response.get('items', []):
+                    subnets.append(GCESubnet(self.provider, subnet))
+                if 'nextPageToken' not in response:
+                    break
+                token = response['nextPageToken']
```
nuwang: @chiniforooshan Another line 431 candidate?

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Done

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Done

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Added the <limit> and <marker> parameters.

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Done

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -872,3 +1071,97 @@ def create(self, name, volume, description=None):
...
+        buckets = []
+        token = None
+        while True:
+            list_result = self.list(marker=token)
+            for bucket in list_result:
+                if name in bucket.name:
+                    buckets.append(bucket)
+            if not list_result.is_truncated:
+                break
+            token = list_result.marker
```
chiniforooshan: Nice! Done

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Good idea :)

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Done

nuwang:
@chiniforooshan That was really fast :-) Awesome.

PR #52, opened 2017-07-31
Fixes for GCE provider issues

machristie:
This pull request fixes some issues that I ran into when testing the GCE integration with the Django cloudlaunch backend:
* cloudlaunch expects that CloudProviders have a `cloud_type` property so I added that
  * perhaps cloud_type should explicitly be added to the interface?
* cloudlaunch REST hyperlink building didn't like putting the selfLink as the image id in the reversed URL.  Looking at the API GCE actually uses the image name as the identifier so I changed GCEMachineImage.id to return the GCE image name.  Besides now working this seems more consistent with how cloudbridge/cloudlaunch expect to use a GCEMachineImage.id.
* in services.py there was one type and then one spot in the code where it assumed that 'items' would be in the response but that's not the case if there are no instances.

### DIFF
--- a/cloudbridge/cloud/providers/aws/provider.py
+++ b/cloudbridge/cloud/providers/aws/provider.py
@@ -31,7 +31,6 @@ class AWSCloudProvider(BaseCloudProvider):
 
     def __init__(self, config):
         super(AWSCloudProvider, self).__init__(config)
-        self.cloud_type = 'aws'
 
         # Initialize cloud connection fields
         self.a_key = self._get_config_value(
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -136,6 +136,7 @@ class GCECloudProvider(BaseCloudProvider):
 
     def __init__(self, config):
         super(GCECloudProvider, self).__init__(config)
+        self.cloud_type = 'gce'
 
         # Initialize cloud connection fields
         self.client_email = self._get_config_value(
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -136,7 +136,6 @@ class GCECloudProvider(BaseCloudProvider):
 
     def __init__(self, config):
         super(GCECloudProvider, self).__init__(config)
-        self.cloud_type = 'gce'
 
         # Initialize cloud connection fields
         self.client_email = self._get_config_value(
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -655,7 +655,7 @@ class GCEMachineImage(BaseMachineImage):
         :rtype: ``str``
         :return: ID for this instance as returned by the cloud middleware.
         """
-        return self._gce_image.get('selfLink')
+        return self._gce_image['name']
 
     @property
     def name(self):
--- a/cloudbridge/cloud/providers/gce/services.py
+++ b/cloudbridge/cloud/providers/gce/services.py
@@ -377,7 +377,7 @@ class GCEImageService(BaseImageService):
             for project in GCEImageService._PUBLIC_IMAGE_PROJECTS:
                 for image in helpers.iter_all(
                         self.provider.gce_compute.images(), project=project):
-                    self._public_iamges.append(
+                    self._public_images.append(
                         GCEMachineImage(self.provider, image))
         except googleapiclient.errors.HttpError as http_error:
                 cb.log.warning("googleapiclient.errors.HttpError: {0}".format(
@@ -548,8 +548,10 @@ class GCEInstanceService(BaseInstanceService):
                               maxResults=max_result,
                               pageToken=marker)
                         .execute())
-        instances = [GCEInstance(self.provider, inst)
-                     for inst in response['items']]
+        instances = []
+        if 'items' in response:
+            instances = [GCEInstance(self.provider, inst)
+                         for inst in response['items']]
         if len(instances) > max_result:
             cb.log.warning('Expected at most %d results; got %d',
                            max_result, len(instances))
--- a/cloudbridge/cloud/providers/openstack/provider.py
+++ b/cloudbridge/cloud/providers/openstack/provider.py
@@ -33,7 +33,6 @@ class OpenStackCloudProvider(BaseCloudProvider):
 
     def __init__(self, config):
         super(OpenStackCloudProvider, self).__init__(config)
-        self.cloud_type = 'openstack'
 
         # Initialize cloud connection fields
         self.username = self._get_config_value(


nuwang:
@machristie Thanks for these changes Marcus. To take each issue in turn:

1. Cloud_type property - I think the cloud_type property needs to change. For one thing, it's a string, so any typo (e.g. cloud_type == "opensack") could cause possibly undetected breakage. Secondly, it's already quite repetitive, since that information is available elsewhere (e.g. in PROVIDER_ID and as isinstance(cloud, GCEProvider) ). My thinking is that perhaps the thing to do is to change all cloud_type == "openstack" type comparisons to cloud.PROVIDER_ID == ProviderList.OPENSTACK. That way, we have a type checked, non-repetitive solution.

2. I agree that this is a significant issue, and probably applies to the whole of the GCE provider, so this is a question that @mbookman @chiniforooshan @baizhang should probably weigh in on. If we use hyperlinks as IDs we will be unable to embed it in a ReST url, as cloudlaunch does, unless it's encoded. I'm not sure why  it's not encoded already though, because it ought happen automatically. Seems dangerous to have unsanitised stuff in there. Nevertheless, it would make for a more pleasing URL if the ID could be a number or a string. 

3. We should probably do a check across the whole of the GCE provider to make sure that any references to response['items'] also checks whether the items entry exists.

machristie:
@nuwang I went ahead and committed a fix for the first issue. Sorry I didn't read further to see you wanted this as a separate bug, but the fix was pretty small.

nuwang:
@machristie That's great, thanks for that. I guess we may need an issue on the cloudlaunch side.

machristie:
Regarding the second issue, here is the exception I get in cloudlaunch when the GCEMachineImage id is the selfLink:

```
[2017-07-31 13:39:36,573 django.request:284 ERROR] Internal Server Error: /api/v1/infrastructure/clouds/us-central1-a/compute/machine_images/
Traceback (most recent call last):
  File "/Users/machrist/SGCI/cloudlaunch/launcher/venv/lib/python3.6/site-packages/rest_framework/relations.py", line 337, in to_representation
    url = self.get_url(value, self.view_name, request, format)
  File "/Users/machrist/SGCI/cloudlaunch/launcher/cloudlaunch/django-cloudlaunch/baselaunch/drf_helpers.py", line 130, in get_url
    raise e
  File "/Users/machrist/SGCI/cloudlaunch/launcher/cloudlaunch/django-cloudlaunch/baselaunch/drf_helpers.py", line 125, in get_url
    view_name, kwargs=reverse_kwargs, request=request, format=format)
  File "/Users/machrist/SGCI/cloudlaunch/launcher/venv/lib/python3.6/site-packages/rest_framework/reverse.py", line 50, in reverse
    url = _reverse(viewname, args, kwargs, request, format, **extra)
  File "/Users/machrist/SGCI/cloudlaunch/launcher/venv/lib/python3.6/site-packages/rest_framework/reverse.py", line 63, in _reverse
    url = django_reverse(viewname, args=args, kwargs=kwargs, **extra)
  File "/Users/machrist/SGCI/cloudlaunch/launcher/venv/lib/python3.6/site-packages/django/core/urlresolvers.py", line 600, in reverse
    return force_text(iri_to_uri(resolver._reverse_with_prefix(view, prefix, *args, **kwargs)))
  File "/Users/machrist/SGCI/cloudlaunch/launcher/venv/lib/python3.6/site-packages/django/core/urlresolvers.py", line 508, in _reverse_with_prefix
    (lookup_view_s, args, kwargs, len(patterns), patterns))
django.core.urlresolvers.NoReverseMatch: Reverse for 'machine_image-detail' with arguments '()' and keyword arguments '{'cloud_pk': 'us-central1-a', 'pk': 'https://www.googleapis.com/compute/v1/projects/centos-cloud/global/images/centos-6-v20131120'}' not found. 1 pattern(s) tried: ['api/v1/infrastructure/clouds/(?P<cloud_pk>[^/.]+)/compute/machine_images/(?P<pk>[^/.]+)/$']
```

This might be an issue with the regular expression used since it is matching on characters that aren't forward slash or period.

But besides this problem, as I mentioned yesterday when talking with @nuwang and @afgane, if you use cloudbridge to get a GCEMachineImage and store the id and then later use cloudbridge to retrieve the GCEMachineImage instance using that id you would use 
```
provider.compute.images.get(image_id)
```
but here the image_id must be the GCE image name since that is what the GCE API uses as an image identifier. See [Google Compute API documentation](https://cloud.google.com/compute/docs/reference/latest/images/get).

So having the GCEMachineImage.id be the GCE image name makes the API self-consistent.

machristie:
@nuwang Yes, I have that fixed in cloudlaunch. I'll open another pull request.  I only found one place where the `cloud_type` attribute was used so it was an easy fix.

mbookman:
I didn't completely follow the issue with the selfLink, but I can say more generally that this has been a problem in the past where the GCE APIs expect tokenized fields (name, zone, project) instead of the selfLink.

That said, I am concerned that if we use the "name" alone as the "id", we (may) lose the ability to reference images from different projects, including "public" or "shared" images. Is there an alternative way to do this where the "id" is the fully qualified selfLink, carrying along the project information?

nuwang:
@mbookman To elaborate further, Marcus is working on adding GCE support to CloudLaunch. CloudLaunch exposes most CloudBridge resources as ReST resources. For example, a URL to machine_images on the OpenStack NeCTAR cloud would look like the following:
```
https://beta.launch.usegalaxy.org/cloudlaunch/api/v1/infrastructure/clouds/nectar/compute/machine_images/b0003c49-486b-491a-b26d-f8fbe6627415/
```
The generalized structure is: 
```
https://beta.launch.usegalaxy.org/cloudlaunch/api/v1/infrastructure/clouds/<cloud_id>/compute/machine_images/<image_id>
```
In GCE, if we make image_id a URL, there would be an (encoded) URL within a ReST URL. Not a showstopper but not all that nice visually. However, I think your point about capturing the project_id remains, so I think we may have to live with that/make it work.

The second issue Marcus highlighted is probably caused by a bug. Assuming that selfLink is the id in cloudbridge, then, to retrieve a particular image:
```
provider.compute.images.get(selfLink)
```
should work, since the get() method accepts an ID.

However, the GCE implementation here: https://github.com/gvlproject/cloudbridge/blob/gce/cloudbridge/cloud/providers/gce/services.py#L395

looks like this:
```
                         gce_compute().images.get(project=self.provider.project_name,
                              image=image_id)
```

where image_id is the selfLink we pass into the get() method.
This probably won't work, since it expects the image_name and not the selfLink. (or is there a way to fetch the image directly by selfLink?) So it looks like the get method would need to extract the project_name and image_name from the selfLink and send those in instead.

Also, I'm guessing this is only a problem for images, and not any of the other resources?

mbookman:
Thanks @nuwang for the details.

Testing [GCE beta API](https://cloud.google.com/compute/docs/reference/beta/) - there doesn't seem to be any indicatation that this is changing (the acceptance of a selfLink for _get_ requests).

In other contexts I have needed to return the selfLink as the "id" and then tokenize it on "image.get" requests. Often the implementation is along the lines of:

```
def get(resource):
  if resource.startswith('https:'):
    # tokenize it into project, resource name
  else:
    # use default project, assume "resource" is the name

  # call the API's get method with project, resource name
```

It sounds like we need to do the same here. If there were a significant reason to, we could truncate the "https://www.googleapis.com" from the selfLink and accept a "resource" that just contains the path, but I think we'd only have slightly less ugly URLs embedded in URLs then.

chiniforooshan:
If I understand this correctly, we already have helper functions for this. So, the get method can simply be something like:
```
def get(self, image_id):
    return GCEMachineImage(self.provider, self.provider.parse_url(image_id).get())
```

nuwang:
@mbookman Thanks for clearing that up - I'm guessing we'll have to go for the latter option of tokenizing it, based on Ehsan's input.
@chiniforooshan I went through the parse_url method but I don't have a full understanding of how it works - I think it would be useful to have more documentation, and some examples, for the GCPResourceUrl and GCPResources objects. If I understood it correctly, the parse_url function can can take a GCE resourceURL, and then extract the variable parts into a dictionary - is that correct? In that case, it sounds like a matter of changing the code to:
```
def get(self, imageId):
   # image_id = selfLink
   elems = self.provider.parse_url(imageId).get()
   gce_compute().images.get(project=elems['project'], image=elems['image'])
```
Does that sound right?

machristie:
Since we decided with the GCE team to take a different approach to the selfLink issue, I'm going to create a new pull request that removes the name/selfLink changes.

machristie:
Note: the non-selfLink changes in this PR have been split into separate PRs: #61 and #62.

PR #58, opened 2017-08-30
Edited SVG image via Adobe Illustrator

malloryfreeberg:
Edited SVG cloudbridge image using Adobe Illustrator and updated the associated raw text file here. Added a box around Network and changed the colors of some boxes.


afgane:
Thanks!

PR #61, opened 2017-08-31
Removing cloud_type attribute: use PROVIDER_ID instead

machristie:
This commit was originally submitted in #52 along with some other commits that are no longer wanted. So this pull request just splits it out into its own PR.

See also [Nuwan's comment on PR #52](https://github.com/gvlproject/cloudbridge/pull/52#issuecomment-319359191) for motivation for this change.

Also note that cloudlaunch has already been updated to remove references to provider cloud_type.


afgane:
Thanks for pulling this out.

Issue #32, opened 2017-04-04
GCE provider issues


afgane:
In testing PR #31, I came across a few issues with the GCE provider so will summarize those here to keep track of them (@chiniforooshan, @baizhang, @mbookman):

- [x]  `KeyError: 'methods'` when creating a provider
Cannot create a GCE provider because getting the following exception:
```
/Users/eafgan/projects/pprojects/cloudbridge-branches/gce/cloudbridge/cloud/providers/gce/provider.pyc in __init__(self, connection)
     83         # We will not mutate self._desc; it's OK to use items() in Python 2.x.
     84         for resource, resource_desc in desc['resources'].items():
---> 85             methods = resource_desc['methods']
     86             if 'get' not in methods:
     87                 continue

KeyError: 'methods'
```
This happens on both py27 and py36 with a provider config like this:
```
gce_config = {'gce_project_name': 'project name',
                       'gce_service_creds_file': 'service acct file.json'}
```
Simply changing line 85 in gce `provider.py` to `methods = resource_desc.get('methods', "")` resolves this but I'm not sure if that's the intended path.

- [x] `TypeError: Missing required parameter "zone"` when operating on some compute services
When trying to run the following commands (possibly others) `provider.compute.instances.list()` or `provider.compute.instance_types.list()`, getting the following traceback:
```
/Users/eafgan/.pyenv/versions/cb-py2/lib/python2.7/site-packages/googleapiclient/discovery.pyc in method(self, **kwargs)
    725     for name in parameters.required_params:
    726       if name not in kwargs:
--> 727         raise TypeError('Missing required parameter "%s"' % name)
    728
    729     for name, regex in six.iteritems(parameters.pattern_params):

TypeError: Missing required parameter "zone"
```
Explicitly defining `gce_default_zone` in the provider config solves this but cannot we have a default default?

- [x] `instances.list` method does not work when no instances are present
Just running `provide.compute.instances.list()` comes back with this when there are no live instances currently available:
```
/Users/eafgan/projects/pprojects/cloudbridge-branches/gce/cloudbridge/cloud/providers/gce/services.pyc in list(self, limit, marker)
    524             pageToken=marker).execute()
    525         instances = [GCEInstance(self.provider, inst)
--> 526                      for inst in response['items']]
    527         return ServerPagedResultList(len(instances) > max_result,
    528                                      response.get('nextPageToken'),

KeyError: 'items'
```

[...]

PR #62, opened 2017-08-31
Fixed typo and handle empty set of instances

machristie:
A couple small fixes for the GCE provider.


afgane:
Thanks. This also fixes one of the bugs reported in #32.

PR #66, opened 2017-09-01
Add image_id and create_image methods to GCEInstance.

baizhang:

This pull request adds image_id and create_image methods to GCEInstance. It also adds some method / method placeholders to reflect new changes in interfaces.

### DIFF
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -142,7 +142,7 @@ class GCECloudProvider(BaseCloudProvider):
             'gce_service_creds_file', os.environ.get('GCE_SERVICE_CREDS_FILE'))
         self.credentials_dict = self._get_config_value(
             'gce_service_creds_dict',
-            json.loads(os.environ.get('GCE_SERVICE_CREDS_DICT') or {}))
+            json.loads(os.environ.get('GCE_SERVICE_CREDS_DICT') or '{}'))
         # If 'gce_service_creds_dict' is not passed in from config and
         # self.credentials_file is available, read and parse the json file to
         # self.credentials_dict.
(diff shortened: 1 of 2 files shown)


cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -142,7 +142,7 @@ def __init__(self, config):
             'gce_service_creds_file', os.environ.get('GCE_SERVICE_CREDS_FILE'))
         self.credentials_dict = self._get_config_value(
             'gce_service_creds_dict',
-            json.loads(os.environ.get('GCE_SERVICE_CREDS_DICT') or {}))
+            json.loads(os.environ.get('GCE_SERVICE_CREDS_DICT') or '{}'))
         # If 'gce_service_creds_dict' is not passed in from config and
```
mbookman: [os.getenv()](https://docs.python.org/2/library/os.html#os.getenv) allows for a default:

`json.loads(os.getenv('GCE_SERVICE_CREDS_DICT', '{}'))`

baizhang: Done.

Thanks for review!

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1466,7 +1508,12 @@ def zone_id(self):
 
     @property
     def source(self):
-        return self._volume.get('sourceSnapshot')
+        return (self._volume.get('sourceSnapshot') or
```
nuwang: This should return a CloudBridge Snapshot or Image object, instead of the unwrapped version.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1466,7 +1508,12 @@ def zone_id(self):
...
+                self._volume.get('sourceImage'))
+
+    @property
+    def source_image(self):
```
nuwang: We don't have this at the interface level, is this needed?

baizhang:
@nuwang thanks for your review. I made changes to address your comments. PTAL. Thanks!

nuwang:
@baizhang Thanks for these changes, looks good.

PR #67, opened 2017-09-01
Implement subnetwork.get_or_create_default

chiniforooshan:
@nuwang @afgane @baizhang @mbookman

### DIFF
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -1384,7 +1384,7 @@ class GCESubnet(BaseSubnet):
 
     @property
     def zone(self):
-        raise NotImplementedError('To be implemented')
+        raise NotImplementedError('GCP subnetworks are regional resources')
 
     @property
     def delete(self):
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -896,10 +896,8 @@ class GCEInstance(BaseInstance):
             return None
         for disk in self._gce_instance['disks']:
             if 'boot' in disk and disk['boot']:
-                boot_disk = GCEVolume(
-                    self._provider,
-                    self._provider.get_gce_resource_data(disk['source']))
-                return boot_disk.source_image
+                return self._provider.get_gce_resource_data(
+                    disk['source']).get('sourceImage')
         return None
 
     @property
@@ -958,12 +956,10 @@ class GCEInstance(BaseInstance):
             return
         for disk in self._gce_instance['disks']:
             if 'boot' in disk and disk['boot']:
-                boot_disk = GCEVolume(
-                    self._provider,
-                    self._provider.get_gce_resource_data(disk['source']))
                 image_body = {
                     'name': name,
-                    'sourceDisk': boot_disk.id
+                    'sourceDisk': self._provider.get_gce_resource_data(
+                        disk['source']).get('selfLink')
                 }
                 operation = (self._provider
                              .gce_compute
@@ -1426,7 +1422,7 @@ class GCESubnet(BaseSubnet):
 
     @property
     def zone(self):
-        raise NotImplementedError('GCP subnetworks are regional resources')
+        return None
 
     @property
     def delete(self):
@@ -1508,12 +1504,15 @@ class GCEVolume(BaseVolume):
 
     @property
     def source(self):
-        return (self._volume.get('sourceSnapshot') or
-                self._volume.get('sourceImage'))
-
-    @property
-    def source_image(self):
-        return self._volume.get('sourceImage')
+        if 'sourceSnapshot' in self._volume:
+            return GCESnapshot(self._provider,
+                               self._provider.get_gce_resource_data(
+                                   self._volume.get('sourceSnapshot')))
+        if 'sourceImage' in self._volume:
+            return GCEMachineImage(self._provider,
+                                   self._provider.get_gce_resource_data(
+                                       self._volume.get('sourceImage')))
+        return None
 
     @property
     def attachments(self):
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -5,6 +5,7 @@ import hashlib
 import inspect
 import io
 import json
+import math
 import re
 import uuid
 
@@ -675,6 +676,16 @@ class GCEMachineImage(BaseMachineImage):
         """
         return self._gce_image.get('description', '')
 
+    @property
+    def min_disk(self):
+        """
+        Returns the minimum size of the disk that's required to
+        boot this image (in GB)
+        :rtype: ``int``
+        :return: The minimum disk size needed by this image
+        """
+        return int(math.ceil(float(self._gce_image.get('diskSizeGb'))))
+
     def delete(self):
         """
         Delete this image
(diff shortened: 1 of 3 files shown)


cloudbridge/cloud/providers/gce/services.py
```diff
@@ -774,10 +773,13 @@ def list(self, network=None, region=None, limit=None, marker=None):
                                      response.get('nextPageToken'),
                                      False, data=subnets)
 
-    def create(self, network, cidr_block, name=None, zone=None):
+    def create(self, network, cidr_block, name=None, zone=None, region=None):
         if not name:
             name = 'subnet-{0}'.format(uuid.uuid4())
-        region = self.provider.region_name
+        if region is not None and zone is not None:
```
mbookman: Does this require explicitly checking for "None" or can it be:

`if region and zone:
`

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -774,10 +773,13 @@ def list(self, network=None, region=None, limit=None, marker=None):
                                      response.get('nextPageToken'),
                                      False, data=subnets)
 
-    def create(self, network, cidr_block, name=None, zone=None):
+    def create(self, network, cidr_block, name=None, zone=None, region=None):
```
mbookman: Could you add a comment explaining how regions and zones are intended to work here.
I'm not sure if the implementation reflects what had been discussed.

I think the intended implementation is:

```
# GCE subnets are regional. The logic for determining the subnet region here is:
# - If a region is provided, use it.
# - If a zone is provided, use the region that it belongs to.
# - If neither is provided, use the default region as set in the provider.
```

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -774,10 +773,13 @@ def list(self, network=None, region=None, limit=None, marker=None):
...
         if not name:
             name = 'subnet-{0}'.format(uuid.uuid4())
-        region = self.provider.region_name
+        if region is not None and zone is not None:
+            cb.log.warning('Both parameters "zone" and "region" are provided')
```
mbookman: Can the message also specify what the outcome is?
Something like:

`Both parameters zone and region are provided; using region.`

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -774,10 +773,13 @@ def list(self, network=None, region=None, limit=None, marker=None):
...
+        if region is None:
+            region = self.provider.region_name
         if isinstance(zone, GCEPlacementZone):
```
mbookman: As I read this, if a zone is specified it ends up being the thing used, even if a region was specified as well.
Is that what was intended?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -797,14 +799,28 @@ def create(self, network, cidr_block, name=None, zone=None):
...
-                cb.log.warning('subnet ID: %s', subnet.id)
                 if subnet.id == response['targetId']:
                     return subnet
         except:
             return None
 
     def get_or_create_default(self, zone=None):
-        raise NotImplementedError('To be implemented')
+        """
+        Every GCP project comes with a default auto mode VPC network. An auto
```
mbookman: I think this is true for every *new* GCP project, but is not true for old projects.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1384,7 +1384,7 @@ def region(self):
 
     @property
     def zone(self):
-        raise NotImplementedError('To be implemented')
+        raise NotImplementedError('GCP subnetworks are regional resources')
```
nuwang: The zone is an optional parameter mainly for informational purposes so it's probably better to return None here.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -797,14 +799,28 @@ def create(self, network, cidr_block, name=None, zone=None):
...
+        mode VPC network has exactly one subnetwork per region. This method
+        returns the subnetwork of the default network that spans the given
+        zone.
+        """
+        network = self.provider.network.get_by_name(
```
nuwang: This should probably be replaced by self.provider.networking.networks.find(name=GCEFirewallsDelegate.DEFAULT_NETWORK)[0]

Also, in some recent refactoring we did, networks have been moved under networking for greater consistency, and the parameter order for network.create() has been changed.

Which also brings up the question, do we need the get_by_name function since the find function exists?

cloudbridge/cloud/providers/gce/services.py
afgane: The same situation happens on AWS so if a provider-default does not exist, we create a CloudBridge-default: https://github.com/gvlproject/cloudbridge/blob/master/cloudbridge/cloud/providers/aws/services.py#L884

cloudbridge/cloud/providers/gce/services.py
afgane: Also, `region` is not a method param in the interface (https://github.com/gvlproject/cloudbridge/blob/master/cloudbridge/cloud/interfaces/services.py#L728). Is it not possible to obtain the region from the zone object? and use the _default_ region when the zone is not provided?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -797,14 +812,27 @@ def create(self, network, cidr_block, name=None, zone=None):
...
+        """
+        Every GCP project comes with a default auto mode VPC network. An auto
+        mode VPC network has exactly one subnetwork per region. This method
+        returns the subnetwork of the default network that spans the given
+        zone.
+        """
+        network = self.provider.network.get_or_create_default()
+        subnets = self.list(network, zone)
+        if len(subnets) > 1:
+            cb.log.warning('The default network has more than one subnetwork '
```
afgane: This seems to go against the intent of this method. As a user, by calling `get_or_create_default` I'd expect to get something back. I guess it's not likely that this condition will be triggered as it's likely there will be only one subnet for the default network so it's probably ok but I wonder if perhaps we should instead just return a random  subnet here as well?

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: N/A: this part is removed in the newer version #74.

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: N/A: this part is removed in the newer version #74.

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Done: fixed in #74.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -626,28 +626,27 @@ def list(self, limit=None, marker=None, filter=None):
         except:
             return []
 
-    def create(self, name):
+    def _create(self, name, autoCreateSubnetworks):
```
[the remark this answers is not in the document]
chiniforooshan: Done: fixed in #74.

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Note that this line just logs a warning. Next lines will actually return something.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Done: fixed in #74.

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: N/A: this part is removed in the newer version #74.

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: N/A: this part is removed in the newer version #74.

cloudbridge/cloud/providers/gce/services.py
chiniforooshan: Done: fixed in #74.

afgane:
@chiniforooshan should we close this PR given #74? There's a couple more bits here but not sure if those are a result of the the git mixup...

cloudbridge/cloud/providers/gce/services.py
afgane: Ahh, yes, I see. I just took the next `if` as `else if`...

PR #68, opened 2017-09-04
Conditional builds and coverage reports

nuwang:
This PR adds support for conditional builds and coverage reports. This way, if a push or pull request affects only a specific set of providers, only those providers' builds will be run. Also, builds are skipped altogether if only documentation has changed. Any changes that occur outside of the providers folder or docs folder will trigger a full build however.  Cron jobs or API based build triggers will also result in a full build.

Coverage reports are skipped altogether if only providers have been changed. The assumption being that test coverage is affected only by interface or test level changes, and less affected by individual provider changes. (This is also needed because coverage will drop drastically when only one provider is tested)

The intent is to save time and costs associated with repeated builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,29 +22,54 @@ matrix:
       env: TOX_ENV=pypy-openstack
 before_install:
     - |
-      DOCS_REGEX='(\.rst$)|(^(docs))/'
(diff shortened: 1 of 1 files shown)


afgane:
I think this is a great change to the testing infrastructure: skipping steps on OpenStack will help avoid failures causes by the infrastructure instability and for commercial providers it will help save $$.

PR #71, opened 2017-09-07
Added tests for resource name validation during resource creation

nuwang:


cloudbridge/cloud/base/resources.py
```diff
@@ -803,7 +831,7 @@ def __repr__(self):
 class BaseNetwork(BaseCloudResource, BaseObjectLifeCycleMixin, Network):
 
     CB_DEFAULT_NETWORK_NAME = os.environ.get('CB_DEFAULT_NETWORK_NAME',
-                                             'CloudBridgeNet')
+                                             'cloudbridge_net')
```
afgane: How about swapping the underscores here to dashes for improved url structure in cases this is used as a resource ID?

nuwang: @afgane Are you thinking cloud-bridge-net, or cloudbridge-net?

afgane: `cloudbridge-net`

nuwang: Thanks for pointing this out. Fixed in: https://mail.google.com/mail/u/0/#inbox/15e5e14123de6359

nuwang: Thanks for pointing this out. Fixed in: https://github.com/gvlproject/cloudbridge/commit/bccac478d8a16f249d34a4cd2b0c87bd24124416

PR #72, opened 2017-09-10
Boto3 upgrade

nuwang:


nuwang:
@01000101 Thanks for all these excellent changes. The idea of using the EC2ServiceFilter is neat, since it simplifies the code quite a bit. I've merged the current changes with the latest master and have started fixing up the remaining issues. We are hoping to make a quick push to upgrade to Boto3. Any input you have would be much appreciated.

01000101:
@nuwang I'm really glad someone was able to pick up where I left off! I'm too busy for the next week (maybe 2) but as soon as I can, I'll check out the updates and see what can be improved. Cheers

PR #73, opened 2017-09-13
Document how to fix macOS/Python 3.6 certificate issue

machristie:
This pull request documents how I fixed an SSL issue that seems like it will likely hit everyone running Python 3.6+ on macOS.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -12,28 +12,6 @@ run::
 
     pip install --upgrade cloudbridge
 
-Common Setup Issues
--------------------
-
-**macOS Issues**
-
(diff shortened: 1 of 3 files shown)


machristie:
Sounds good @nuwang, will do

machristie:
@nuwang I've moved the documentation to a separate page. Please review when you have a chance.

afgane:
Thanks for the update. I'll merge this but then rename the page from _Common setup issues_ to _Troubleshooting_ because we don't commonly see issues ;)

machristie:
Makes sense 😄 Thanks!

PR #77, opened 2017-09-19
Boto3 upgrade

nuwang:
This PR completes the boto3 upgrade work originally started by @01000101.

All tests pass against both mocks and AWS. One significant issue is that the implementation of server side paging was made considerably complex because at present, Boto3 doesn't provide a way to obtain a resumption token when listing Resources. This means that we end up having to choose between the following two:
1. Use Resources but give up on server paging/resort to client paging
2. Use server paging but give up on Resources and all the goodies that come with it, dealing with json dicts instead.

Since neither option is desirable, I've hacked around it for now by directly accessing some protected members in Boto3 - which is again undesirable, but allows us to use Resources with server paging. I've also filed this issue: https://github.com/boto/boto3/issues/1268 requesting a fix.

In a worst case scenario, possibly the easiest thing to do is to simply fall back to client paging and continue to use boto Resources, at the cost of some performance, but with much much simpler code.

### DIFF
--- a/cloudbridge/cloud/providers/aws/helpers.py
+++ b/cloudbridge/cloud/providers/aws/helpers.py
@@ -32,7 +32,7 @@ def trim_empty_params(params_dict):
 class BotoGenericService(object):
     '''
     Generic implementation of a Boto3 AWS service. Uses Boto3
-    resource, collection and paging primitives to implement
+    resource, collection and paging support to implement
     basic cloudbridge methods.
 
     :param AWSCloudProvider provider: CloudBridge AWS provider
@@ -139,18 +139,18 @@ class BotoGenericService(object):
         client = self.boto_conn.meta.client
         list_op = self._get_list_operation()
         paginator = client.get_paginator(list_op)
+        PaginationConfig = {}
         if limit:
-            PaginationConfig = {
-                'MaxItems': limit, 'PageSize': limit}
-        else:
-            PaginationConfig = None
-        params.update({'NextToken': marker,
-                       'PaginationConfig': PaginationConfig})
+            PaginationConfig = {'MaxItems': limit, 'PageSize': limit}
+        if marker:
+            PaginationConfig.update({'StartingToken': marker})
+        params.update({'PaginationConfig': PaginationConfig})
         args = trim_empty_params(params)
         pages = paginator.paginate(**args)
+        # resume_token is not populated unless the iterator is used
+        items = list(self._resource_iterator(collection, params, pages, limit))
         resume_token = pages.resume_token
-        return (resume_token,
-                self._resource_iterator(collection, params, pages, limit))
+        return (resume_token, items)
 
     def _make_query(self, collection, limit, marker):
         '''
@@ -178,9 +178,11 @@ class BotoGenericService(object):
         results = [self.cb_resource(self.provider, obj) for obj in boto_objs]
 
         if resume_token:
-            log.debug("Received a resume token, implying server pagination.")
-            return ServerPagedResultList(self.provider, results,
-                                         limit=limit, marker=resume_token)
+            log.debug("Received a resume token, using server pagination.")
+            return ServerPagedResultList(is_truncated=True,
+                                         marker=resume_token,
+                                         supports_total=False,
+                                         data=results)
         else:
             log.debug("Did not received a resume token, will page in client"
                       " if necessary.")
--- a/cloudbridge/cloud/providers/aws/helpers.py
+++ b/cloudbridge/cloud/providers/aws/helpers.py
@@ -10,7 +10,7 @@ from cloudbridge.cloud.base.resources import ServerPagedResultList
 
 
 def trim_empty_params(params_dict):
-    '''
+    """
     Given a dict containing potentially null values, trims out
     all the null values. This is to please Boto, which throws
     a parameter validation exception for NoneType arguments.
@@ -25,23 +25,48 @@ def trim_empty_params(params_dict):
             'GroupName': 'abc',
             'VpcId': 'xyz',
         }
-    '''
+    """
     return {k: v for k, v in params_dict.items() if v is not None}
 
 
+def find_tag_value(tags, key):
+    """
+    Finds the value associated with a given key from a list of AWS tags.
+
+    :type tags: list of ``dict``
+    :param tags: The AWS tag list to search through
+
+    :type key: ``str``
+    :param key: Name of the tag to search for
+    """
+    for tag in tags or []:
+        if tag.get('Key') == key:
+            return tag.get('Value')
+    return None
+
+
 class BotoGenericService(object):
-    '''
+    """
     Generic implementation of a Boto3 AWS service. Uses Boto3
     resource, collection and paging support to implement
     basic cloudbridge methods.
-
-    :param AWSCloudProvider provider: CloudBridge AWS provider
-    :param CloudResource cb_iface: CloudBridge class to wrap results in
-    :param Boto3.Resource boto_conn: Boto top level service resource
-                                     (e.g. EC2, S3) connection.
-    :param str boto_collection_name: Boto collection name
-    '''
+    """
     def __init__(self, provider, cb_resource, boto_conn, boto_collection_name):
+        """
+        :type provider: :class:`AWSCloudProvider`
+        :param provider: CloudBridge AWS provider to use
+
+        :type cb_resource: :class:`CloudResource`
+        :param cb_resource: CloudBridge Resource class to wrap results in
+
+        :type boto_conn: :class:`Boto3.Resource`
+        :param boto_conn: Boto top level service resource (e.g. EC2, S3)
+                          connection.
+
+        :type boto_collection_name: ``str``
+        :param boto_collection_name: Boto collection name that corresponds
+                                    to the CloudBridge resource (e.g. key_pair)
+        """
         self.provider = provider
         self.cb_resource = cb_resource
         self.boto_conn = boto_conn
@@ -67,12 +92,12 @@ class BotoGenericService(object):
         return getattr(self.boto_conn, resource_model.name)
 
     def get(self, resource_id):
-        '''
+        """
         Returns a single resource.
 
-        :param str resource_id: ID of the boto resource to fetch
-        :returns: CloudBridge object or None
-        '''
+        :type resource_id: ``str``
+        :param resource_id: ID of the boto resource to fetch
+        """
         try:
             log.debug("Retrieving resource: %s with id: %s",
                       self.boto_collection_model.name, resource_id)
@@ -90,30 +115,30 @@ class BotoGenericService(object):
                 raise e
 
     def _get_list_operation(self):
-        '''
+        """
         This function discovers the list operation for a particular resource
         collection. For example, given the resource collection model for
         KeyPair, it returns the list operation for it, as describe_key_pairs.
-        '''
+        """
         return xform_name(self.boto_collection_model.request.operation)
 
     def _to_boto_resource(self, collection, params, page):
-        '''
+        """
         This function duplicates some of the logic of the pages() method in
         boto.resources.collection.ResourceCollection. It will convert a raw
         json response to the corresponding Boto resource. It's necessary
         because paginators() return json responses, and there's no direct way
         to convert a paginated json response to a Boto Resource.
-        '''
+        """
         return collection._handler(collection._parent, params, page)
 
     def _resource_iterator(self, collection, params, pages, limit):
-        '''
+        """
         Iterates through the pages of a paginated result, converting the
         objects to BotoResources as necessary. This duplicates the logic in
         boto's ResourceCollection(). pending issue:
         https://github.com/boto/boto3/issues/1268
-        '''
+        """
         count = 0
         for page in pages:
             for item in self._to_boto_resource(collection, params, page):
@@ -123,12 +148,12 @@ class BotoGenericService(object):
                 yield item
 
     def _get_paginated_results(self, limit, marker, collection):
-        '''
+        """
         If a Boto Paginator is available, use it. The results
         are converted back into BotoResources by directly accessing
         protected members of ResourceCollection. This logic can be removed
         depending on issue: https://github.com/boto/boto3/issues/1268.
-        '''
+        """
         cleaned_params = collection._params.copy()
         cleaned_params.pop('limit', None)
         cleaned_params.pop('page_size', None)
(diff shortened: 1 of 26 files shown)


nuwang:
@afgane @01000101 Please take a look. At this point, I'm pretty jaded with this, so it would be great to make sure nothing has been lost in translation :-)

cloudbridge/cloud/providers/aws/services.py
```diff
@@ -597,103 +409,60 @@ def _process_block_device_mappings(self, launch_config, zone=None):
...
-        reservations = self.provider.ec2_conn.get_all_reservations(
-            max_results=limit,
-            next_token=marker)
-        instances = [AWSInstance(self.provider, inst)
-                     for res in reservations
-                     for inst in res.instances]
-        return ServerPagedResultList(reservations.is_truncated,
-                                     reservations.next_token,
-                                     False, data=instances)
+        return self.iface.list(limit=limit, marker=marker)
```
afgane: This will include all instances (running, stopped, launching shutting down or terminated). Do we want that or should a filter for running ones only?

cloudbridge/cloud/providers/aws/resources.py
```diff
@@ -559,29 +479,28 @@ def id(self):
 
     @property
     def name(self):
-        """
-        Get the snapshot name.
-
-        .. note:: an instance must have a (case sensitive) tag ``Name``
-        """
-        return self._snapshot.tags.get('Name')
+        for tag in self._snapshot.tags or list():
```
afgane: This construct for getting the name keeps repeating. Should it be extracted into a private helper method?

cloudbridge/cloud/providers/aws/resources.py
```diff
@@ -652,79 +560,71 @@ def __init__(self, provider, security_group):
...
+                    continue
+            elif src_group_id:
+                if src_group_id not in [
+                    group_pair.get('GroupId') for group_pair in
+                        rule.get('UserIdGroupPairs', [])]:
+                    continue
+            return AWSSecurityGroupRule(self._provider, rule, self)
         return None
 
     def to_json(self):
```
afgane: We could use the ec2 client `describe_security_groups` method here directly and not worry about translating this into json: http://boto3.readthedocs.io/en/stable/reference/services/ec2.html?highlight=create_volume#EC2.Client.describe_security_groups:
`return self._provider.ec2_conn.meta.client.describe_security_group(GroupIds=[self.id])`

cloudbridge/cloud/providers/aws/resources.py
```diff
@@ -788,91 +688,83 @@ def to_json(self):
...
-        """
-        Delete this object.
-
-        :rtype: bool
-        :return: True if successful
-        """
-        self._key.delete()
+        self._obj.delete()
 
     def generate_url(self, expires_in=0):
```
afgane: Unless I'm looking at the wrong docs, the default expiry value is not valid: http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html

Looks like 1 is min and 604800 (seven days) is max?

cloudbridge/cloud/providers/aws/helpers.py
```diff
@@ -0,0 +1,260 @@
...
+            log.debug("Retrieving resource: %s with id: %s",
+                      self.boto_collection_model.name, resource_id)
+            obj = self.boto_resource(resource_id)
+            obj.load()
+            log.debug("Successfully Retrieved: %s", obj)
+            return self.cb_resource(self.provider, obj)
+        except ClientError as e:
+            error_code = e.response['Error']['Code']
+            if any(status in error_code for status in
+                    ('NotFound', 'InvalidParameterValue', 'Malformed', '404')):
```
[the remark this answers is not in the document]
nuwang: The reason for doing it this way is that the actual error code is of the pattern InvalidAMIID.NotFound, InvalidAMIID.Malformed, InvalidParamterValue etc. etc. so the lookup needs to be reversed.

cloudbridge/cloud/providers/aws/helpers.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def delete(self, resource_id):
+        '''
+        Deletes a resource by id
+
+        :param str id: ID of the resource
+        '''
+        res = self.get(resource_id)
+        if res:
+            res.delete()
```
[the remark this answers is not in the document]
nuwang: I prefer removing the bool for two reasons: One is that it seems redundant to return a bool if delete didn't throw an exception, which it should if the delete didn't succeed I think. The second is that only AWS returns a bool from delete.

cloudbridge/cloud/providers/aws/services.py
```diff
@@ -72,182 +64,63 @@ def __init__(self, provider):
...
-                keynames=[key_pair_id])
-            return AWSKeyPair(self.provider, kps[0])
-        except EC2ResponseError as ec2e:
-            if ec2e.code == 'InvalidKeyPair.NotFound':
-                return None
-            elif ec2e.code == 'InvalidParameterValue':
-                return None
-            else:
-                raise ec2e
+        return self.iface.get(key_pair_id)
```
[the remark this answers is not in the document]
nuwang: I tend to agree - I think svc sounds good.

cloudbridge/cloud/providers/aws/services.py
nuwang: Hmm... Not sure. We should probably list stopped, because we could restart it. Shutting down too probably, since that could fail perhaps? Terminated  is probably ok either way because our handling for this is not too bad at the moment - because we check for both terminated, and error statuses in wait_for etc.

cloudbridge/cloud/providers/aws/services.py
```diff
@@ -953,15 +698,19 @@ class AWSGatewayService(BaseGatewayService):
...
+                                    boto_collection_name='internet_gateways')
 
     def get_or_create_inet_gateway(self, name):
         AWSInternetGateway.assert_valid_resource_name(name)
 
-        gateway = self.provider.vpc_conn.create_internet_gateway()
-        cb_gateway = AWSInternetGateway(self.provider, gateway)
-        cb_gateway.wait_till_ready()
+        cb_gateway = self.iface.create('create_internet_gateway')
+        # self.iface_igws.wait_for_create(cb_gateway.id, 'internet-gateway-id')
```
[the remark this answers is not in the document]
nuwang: Should be removed. Will do.

cloudbridge/cloud/providers/aws/resources.py
```diff
@@ -240,177 +207,138 @@ def __init__(self, provider, ec2_instance):
...
-        """
-        Refreshes the state of this instance by re-querying the cloud provider
-        for its latest state.
-        """
         try:
-            self._ec2_instance.update(validate=True)
-        except (EC2ResponseError, ValueError):
+            self._ec2_instance.reload()
+        except ClientError:
             # The volume no longer exists and cannot be refreshed.
```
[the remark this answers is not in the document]
nuwang: Good catch :-)

cloudbridge/cloud/providers/aws/resources.py
```diff
@@ -240,177 +207,138 @@ def __init__(self, provider, ec2_instance):
...
             # set the status to unknown
-            self._ec2_instance.status = 'unknown'
+            self._ec2_instance.state = {'Name': InstanceState.UNKNOWN}
+
+    def wait_till_exists(self, timeout=None, interval=None):
```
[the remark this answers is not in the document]
nuwang: Yes, that makes sense. In general, I feel that our wait states aren't all that great. We should probably have a few more for common scenarios, like boto3 does.

cloudbridge/cloud/providers/aws/resources.py
nuwang: Good point. I'd been planning to do that and forgot.

cloudbridge/cloud/providers/aws/resources.py
nuwang: Can you elaborate - I didn't quite follow. I'm hoping this part can be redone anyway when we do the interface consistency change for security groups rules, so that might be a good time to address this.

cloudbridge/cloud/providers/aws/resources.py
nuwang: Looks like an error at the interface definition level. We should probably make expires_in default to None, and if it's specified, we need to mandate some default values that apply across all cloud providers.

cloudbridge/cloud/providers/aws/helpers.py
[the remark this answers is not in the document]
afgane: boto3 is inconsistent in what it returns from a delete action. For example: 
- Security Groups delete doesn't return anything: http://boto3.readthedocs.io/en/stable/reference/services/ec2.html#EC2.SecurityGroup.delete
- Instance terminate returns a pretty comprehensive dict: http://boto3.readthedocs.io/en/stable/reference/services/ec2.html#EC2.Instance.terminate
- VPC delete doesn't return anything: http://boto3.readthedocs.io/en/stable/reference/services/ec2.html#EC2.Vpc.delete

Our OpenStack side of the implementation makes a best effort guess, for example: https://github.com/gvlproject/cloudbridge/blob/master/cloudbridge/cloud/providers/openstack/services.py#L876

Perhaps returning a `bool` is filled with peril in the long run because we're taking on perhaps more than necessary but it certainly is nice to get some degree to confirmation in the client from calling a `delete` method vs. having to do `delete` followed by `get` to check if the action succeeded.

nuwang:
Thanks for this very thorough review - I'll make the changes, update and merge unless something else comes up.

cloudbridge/cloud/providers/aws/services.py
afgane: I guess it boils down to consistency across providers. terminated instances won't show up on OpenStack so if they're included here will that mean we'll force clients to filter the results to get consistent behavior? Such filtering is likely required because of instances that may be shutting down or initializing anyhow so maybe including all is the right thing to do and just adding a comment about this...

cloudbridge/cloud/providers/aws/resources.py
afgane: It looks like `boto3` client has the `describe` method for various objects that return the native native object representation so rather than providing our own implementation like we had to with `boto`, it should be easier to just call the native `boto3` method. This is an example of what the describe method will return:
```
[{'Description': 'default group',
   'GroupId': 'sg-7d10cc15',
   'GroupName': 'default',
   'IpPermissions': [{'FromPort': 0,
     'IpProtocol': 'tcp',
     'IpRanges': [],
     'Ipv6Ranges': [],
     'PrefixListIds': [],
     'ToPort': 65535,
     'UserIdGroupPairs': [{'GroupId': 'sg-7d10cc15',
       'GroupName': 'default',
       'UserId': '347162595075'}]},
    {'FromPort': 0,
     'IpProtocol': 'udp',
     'IpRanges': [],
     'Ipv6Ranges': [],
     'PrefixListIds': [],
     'ToPort': 65535,
     'UserIdGroupPairs': [{'GroupId': 'sg-7d10dc15',
       'GroupName': 'default',
       'UserId': '347162595075'}]},
    {'FromPort': -1,
     'IpProtocol': 'icmp',
     'IpRanges': [],
     'Ipv6Ranges': [],
     'PrefixListIds': [],
     'ToPort': -1,
     'UserIdGroupPairs': [{'GroupId': 'sg-7d10dc15',
       'GroupName': 'default',
       'UserId': '9999999999'}]}],
   'IpPermissionsEgress': [],
   'OwnerId': '999999999'}]
```
Granted, it's dict but we could just return `json.loads` of that?

cloudbridge/cloud/providers/aws/resources.py
afgane: It doesn't look like there's max length of time before a URL expires on OpenStack (https://docs.openstack.org/swift/latest/api/temporary_url_middleware.html) so perhaps we should stick to the AWS 7 day max and default to 24 hours? Ping @VJalili

cloudbridge/cloud/providers/aws/helpers.py
```diff
@@ -0,0 +1,260 @@
...
+    Generic implementation of a Boto3 AWS service. Uses Boto3
+    resource, collection and paging support to implement
+    basic cloudbridge methods.
+
+    :param AWSCloudProvider provider: CloudBridge AWS provider
+    :param CloudResource cb_iface: CloudBridge class to wrap results in
+    :param Boto3.Resource boto_conn: Boto top level service resource
+                                     (e.g. EC2, S3) connection.
+    :param str boto_collection_name: Boto collection name
+    '''
```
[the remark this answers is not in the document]
nuwang: Fixed

cloudbridge/cloud/providers/aws/helpers.py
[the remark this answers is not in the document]
nuwang: > Perhaps returning a bool is filled with peril in the long run because we're taking on perhaps more than necessary but it certainly is nice to get some degree to confirmation in the client from calling a delete method vs. having to do delete followed by get to check if the action succeeded.

I think the problem here is that whether confirmation is needed or not is situation specific. There may be cases where, if the provider can't honour the request to delete the object, the user can't do much about it anyway, so they'll just have to continue. Incurring an unnecessary performance hit for a forced check seems spurious in this case, and possibly undesirable.

A good example of this is instance.terminate(), soon to be renamed instance.delete(). We expect users to use `wait_for(InstanceState.DELETED)` in this case anyway. Most often, they may just issue instance.delete(), go do something else, then check back later whether the instance was deleted. So in this case, forcing a wait would be very problematic.

Perhaps we can accept a parameter called, wait=True or verify=True or something, so that the method will block till the delete is complete, or throw an exception. That is, internally do a `wait_for([ObjState.DELETED, ObjState.UNKNOWN]).`

cloudbridge/cloud/providers/aws/services.py
nuwang: Hmm... Interesting point.  Yeah, I guess it'll be quite difficult to achieve 100% consistency in these cases, since multiple provider states are mapped to the same CloudBridge state. Probably docs are the way to go.

cloudbridge/cloud/providers/aws/resources.py
nuwang: So you're saying we should make a server call instead of using the cached ip_permissions value?

cloudbridge/cloud/providers/aws/services.py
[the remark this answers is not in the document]
nuwang: Done

cloudbridge/cloud/providers/aws/resources.py
nuwang: Done

nuwang:
@afgane @01000101 Have made the changes. I guess we'll mainstream this now and we can still fix things as and when they come up. Thanks again for the awesome work you did on the generalised implementation for boto3 @01000101.

cloudbridge/cloud/providers/aws/resources.py
VJalili: I usually set it to 7200 min (and it is mainly because of large genomic files); but I guess 24h as default is reasonable.

PR #82, opened 2017-09-21
Fixes

chiniforooshan:
@baizhang, could you please do the first round of review? Thanks you!

### DIFF
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -840,7 +840,7 @@ class GCEInstance(BaseInstance):
         if machine_type_uri is None:
             return None
         parsed_uri = self._provider.parse_url(machine_type_uri)
-        return GCEInstanceType(self._provider, parsed_uri.get())
+        return GCEInstanceType(self._provider, parsed_uri.get_resource())
 
     def reboot(self):
         """
@@ -896,8 +896,8 @@ class GCEInstance(BaseInstance):
             return None
         for disk in self._gce_instance['disks']:
             if 'boot' in disk and disk['boot']:
-                disk_resource = self._provider.parse_url(disk['source']).get()
-                return disk_resource.get('sourceImage')
+                disk_url = self._provider.parse_url(disk['source'])
+                return disk_url.get_resource().get('sourceImage')
         return None
 
     @property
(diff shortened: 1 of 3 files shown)


baizhang:
LGTM.

Thanks a lot for the fix, Ehsan!

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -896,8 +896,8 @@ def image_id(self):
             return None
         for disk in self._gce_instance['disks']:
             if 'boot' in disk and disk['boot']:
-                return self._provider.get_gce_resource_data(
-                    disk['source']).get('sourceImage')
+                disk_resource = self._provider.parse_url(disk['source']).get()
```
nuwang: The method name get() seems a bit ambiguous when reading the code. It's not immediately obvious what it is that it's retrieving. Should it be something like get_resource() instead?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -828,8 +828,8 @@ def instance_type_id(self):
         machine_type_uri = self._gce_instance.get('machineType')
         if machine_type_uri is None:
             return None
-        instance_type = self._provider.get_gce_resource_data(machine_type_uri)
-        return instance_type.get('name', None)
+        parsed_uri = self._provider.parse_url(machine_type_uri)
```
nuwang: It would be great to have an example url in the GCPResource.parse_url() method docs so you can see what kind of URL is being split apart, and what the split parameters look like.

In addition, an example return dict in the method docs for GCPResourceURL.get() would also be helpful to understand the code without further referencing.

About the TODO item in GCPResources. I can see why there's a need to access the protected member in connection._resourceDesc in GCPResources - it seems a lot more efficient than hitting the endpoint again to retrieve resource definitions. If you think it's unlikely to change in the near future, I think we may as well leave it as is, with the TODO changing to a "FIX_IF_NEEDED". Boto3 has a similar mechanism, however, boto3's resource definitions/metadata is public, so you can access it safely.

chiniforooshan:
Thanks Bai and Nuwan.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Done

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Done

nuwang:
@chiniforooshan Looks great! thanks for that super-fast update.

PR #89, opened 2017-09-28
Refactor firewalls

nuwang:
This PR does the following:
1. Renames security groups->VM firewall
2. Increased the uniformity of SecurityGroupRules by making them service like i.e. have CRUD methods and iteration. However, they have not been turned into a cloudbridge "Service" because it creates circular dependencies, since the firewall must have a reference to the rules. Instead, the VMFirewallRuleContainer remains in resources.py
3. Nova security groups do not support traffic direction. They have also bee deprecated, preventing a nova upgrade. Therefore, they've been replaced by neutron security groups, with a partial migration to the Openstack SDK.

### DIFF
--- a/cloudbridge/cloud/base/resources.py
+++ b/cloudbridge/cloud/base/resources.py
@@ -733,10 +733,11 @@ class BaseVMFirewallRule(BaseCloudResource, VMFirewallRule):
         return self._name
 
     def __repr__(self):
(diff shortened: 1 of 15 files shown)


nuwang:
@afgane I'll merge this in since the next set of changes are dependent on this, and we'll merge this in at some point anyway, but if you can take a look at the PR diff to make sure I haven't missed anything, that'd be great.

Issue #85, opened 2017-09-25
Log everything


afgane:
When debugging or developing, it's useful to get frequent and detailed logging messages. We should thus have log messages (at the `debug` level) at every method start and as appropriate elsewhere. 

This is a rather mechanical task but will be super-useful in the long run!

PR #94, opened 2017-10-05
Add some logging messages - Issue #85

FabioRosado:
Hello, 
I've decided to give a go on this issue but I'm not entirely sure if this is okay.
Could you please let me know if this is something that you want or if I should be adding `logging.debug` to something else?

Thank you!

### DIFF
--- a/cloudbridge/cloud/base/provider.py
+++ b/cloudbridge/cloud/base/provider.py
@@ -1,7 +1,7 @@
 """Base implementation of a provider interface."""
 import functools
-import os
 import logging
+import os
 from os.path import expanduser
 try:
     from configparser import ConfigParser
@@ -39,8 +39,8 @@ class BaseConfiguration(Configuration):
         :rtype: ``int``
         :return: The maximum number of results to return
         """
-        log.debug("Maximum number of results for list methods "
-                  + DEFAULT_RESULT_LIMIT)
+        log.debug("Maximum number of results for list methods %s",
+                  DEFAULT_RESULT_LIMIT)
         return self.get('default_result_limit', DEFAULT_RESULT_LIMIT)
 
     @property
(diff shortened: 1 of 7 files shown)


nuwang:
@FabioRosado This is exactly along the lines of what's needed - many thanks for taking this on.  Ideally, cloudbridge would have log statements at the start of each method detailing the parameters received, and at each control flow statement so that it's possible to know exactly what the library was doing if/when it fails, as well as of course, exceptions.

Having log.debug for most things is fine, but if you spot anything where a log.info or log.warn would do, feel free to change it up, we can always change them around at leisure. log.exception would be preferred for exceptions so there's a detailed stacktrace.

Thanks again for taking this on - this is a critical feature for the plan 1.0.0 release.

cloudbridge/cloud/base/provider.py
```diff
@@ -37,20 +39,26 @@ def default_result_limit(self):
         :rtype: ``int``
         :return: The maximum number of results to return
         """
+        log.debug("Maximum number of results for list methods "
```
nuwang: pylint recommends passing debug parameters as follows:
log.debug("Maximum number of results for list methods %s", DEFAULT_RESULT_LIMIT)

cloudbridge/cloud/base/provider.py
```diff
@@ -94,6 +102,7 @@ def authenticate(self):
             self.security.key_pairs.list()
             return True
         except Exception as e:
+            log.debug("ProviderConnectionException ocurred")
```
nuwang: using log.exception here is recommended, as it will give a detailed stack trace.
```
log.exception("ProviderConnectionException occured" ,e)
```

cloudbridge/cloud/factory.py
```diff
@@ -129,6 +135,7 @@ def create_provider(self, name, config):
             raise NotImplementedError(
                 'A provider with name {0} could not be'
                 ' found'.format(name))
+        log.debug("Provider name: " + name + " config details:" + config)
```
nuwang: Same issue here - pylint recommends parameters instead of concatenation

FabioRosado:
Awesome, I'm glad I was somewhat in the right direction haha 
I've never done something like this before so I'm happy to give a hand! I'll change the logging messages to follow pylint recommendations, thank you for pointing this out.

FabioRosado:
I've followed your suggestions on how to log the messages, fixed the location where I was importing logging so Travis should be happy now.
I still haven't touched the providers but will do as soon as I can 👍

PR #105, opened 2017-11-10
Add support for GCE resource get method to use either id (resource URL) or short name

baizhang:

This pull request includes following changes:
(1) add support for GCE resource get method to use either id (resource URL) or short name;
(2) make it consistent to use resource URL as GCE resource ID (wherever applicable).

Thanks!

### DIFF
--- a/cloudbridge/cloud/providers/gce/provider.py
+++ b/cloudbridge/cloud/providers/gce/provider.py
@@ -8,8 +8,10 @@ import re
 import time
 from string import Template
 
+import cloudbridge as cb
 from cloudbridge.cloud.base import BaseCloudProvider
 
+import googleapiclient
 from googleapiclient import discovery
 
 from oauth2client.client import GoogleCredentials
@@ -55,8 +57,9 @@ class GCPResourceUrl(object):
 
 class GCPResources(object):
 
-    def __init__(self, connection):
+    def __init__(self, connection, provider):
         self._connection = connection
+        self._provider = provider
 
         # Resource descriptions are already pulled into the internal
         # _resourceDesc field of the connection.
@@ -157,6 +160,38 @@ class GCPResources(object):
                 out.parameters[parameter] = m.group(index + 1)
             return out
 
+    def get_resource_url_with_default(self, resource, identifier):
+        """
+        Build a GCPResourceUrl from a resource's name and its identifier.
+        If the identifier is a valid GCP resource URL, then we build the
+        GCPResourceUrl object by parsing this URL. If the identifier is its
+        short name, then we build the GCPResourceUrl object by constructing
+        the resource URL with default project, region, zone values.
+        """
+        # If identifier is a valid GCP resource URL, the parse it.
+        if identifier.startswith(self._root_url):
+            return self.parse_url(identifier)
+        # Otherwise, construct resource URL with default values.
+        if resource not in self._resources:
+            return None
+        resource_desc = self._connection._resourceDesc['resources'][resource]
+        methods = resource_desc.get('methods', {})
+        if 'get' not in methods:
+            cb.log.warning('Resouce ' + resource + ' does not have get method.')
+            return None
+        method = methods['get']
+        default_values = {
+            'project': self._provider.project_name,
+            'region': self._provider.region_name,
+            'zone': self._provider.default_zone,
+        }
+        for param in method['parameterOrder']:
+            if param not in default_values:
+                default_values[param] = identifier
+        resource_url = (self._connection._resourceDesc['baseUrl'] +
+                        method['path'].format(**default_values))
+        return self.parse_url(resource_url)
+
 
 class GCECloudProvider(BaseCloudProvider):
 
@@ -198,8 +233,8 @@ class GCECloudProvider(BaseCloudProvider):
         self._block_store = GCEBlockStoreService(self)
         self._object_store = GCSObjectStoreService(self)
 
-        self._compute_resources = GCPResources(self.gce_compute)
-        self._storage_resources = GCPResources(self.gcp_storage)
+        self._compute_resources = GCPResources(self.gce_compute, self)
+        self._storage_resources = GCPResources(self.gcp_storage, self)
 
     @property
     def compute(self):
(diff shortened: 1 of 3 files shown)


cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -157,6 +160,38 @@ def parse_url(self, url):
...
+        short name, then we build the GCPResourceUrl object by constructing
+        the resource URL with default project, region, zone values.
+        """
+        # If identifier is a valid GCP resource URL, then parse it.
+        if identifier.startswith(self._root_url):
+            return self.parse_url(identifier)
+        # Otherwise, construct resource URL with default values.
+        if resource not in self._resources:
+            return None
+        resource_desc = self._connection._resourceDesc['resources'][resource]
```
chiniforooshan: For the rest of this method I prefer the following slightly simpler and more direct implementation. What do you think?

```
parsed_url = GCPResourceUrl(resource, self._connection)
for _, parameter in self._resources[resource]['parameters']:
  if parameter in DEFAULT_VALUES:
    parsed_url.parameters[parameter] = self._parameter_defaults[parameter]
  else:
    parsed_url.parameters[parameter] = name
return parsed_url
```

Define `self._parameter_defaults` in `__init__` since they are not supposed to be changed.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -157,6 +160,38 @@ def parse_url(self, url):
                 out.parameters[parameter] = m.group(index + 1)
             return out
 
+    def get_resource_url_with_default(self, resource, identifier):
```
chiniforooshan: After this PR resource ID should mean resource URL everywhere and the last parameter of this method should be resource `name`, right?

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -270,3 +305,19 @@ def wait_for_operation(self, operation, region=None, zone=None):
     def parse_url(self, url):
         out = self._compute_resources.parse_url(url)
         return out if out else self._storage_resources.parse_url(url)
+
+    def get_resource(self, resource, identifier):
```
chiniforooshan: `get_resource_by_name` would be more clear. Also `identifier` -> `name`.

cloudbridge/cloud/providers/gce/provider.py
chiniforooshan: Looks like this is never called with a URL. If that's the case maybe something like `resource_name_to_url` or `get_resource_url_by_name` is better?

baizhang: Yes. After this PR,  resource id will be resource URL (GCE selfLink), except where a resource is not mapped to a real GCE service (such as GCESecurityGroup).

baizhang: As discussed in the meeting, I changed the parameter name identifier to url_or_name to make it more clear.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -157,6 +160,38 @@ def parse_url(self, url):
...
+    def get_resource_url_with_default(self, resource, identifier):
+        """
+        Build a GCPResourceUrl from a resource's name and its identifier.
+        If the identifier is a valid GCP resource URL, then we build the
+        GCPResourceUrl object by parsing this URL. If the identifier is its
+        short name, then we build the GCPResourceUrl object by constructing
+        the resource URL with default project, region, zone values.
+        """
+        # If identifier is a valid GCP resource URL, then parse it.
+        if identifier.startswith(self._root_url):
```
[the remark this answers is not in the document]
baizhang: These two lines are still necessary, because it first trims off rootUrl, then these two lines trims off servicePath.

cloudbridge/cloud/providers/gce/provider.py
baizhang: Done. Thank you very much for the suggestion.

cloudbridge/cloud/providers/gce/provider.py
baizhang: As discussed in the meeting, I changed the parameter name identifier to url_or_name to make it more clear.

baizhang:
Hi @chiniforooshan 

Thank you very much for your review! PTAL.

cloudbridge/cloud/providers/gce/provider.py
```diff
@@ -270,3 +301,19 @@ def wait_for_operation(self, operation, region=None, zone=None):
...
+    def get_resource(self, resource, url_or_name):
+        resource_url = (
+            self._compute_resources.get_resource_url_with_default(
+                resource, url_or_name) or
+            self._storage_resources.get_resource_url_with_default(
+                resource, url_or_name))
+        if resource_url is None:
+            return None
+        try:
+            res = resource_url.get_resource()
```
chiniforooshan: nit: could return here and save one line.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -870,14 +834,9 @@ def get(self, volume_id):
...
-        try:
-            return GCEVolume(self.provider,
-                             self.provider.parse_url(volume_id).get_resource())
-        except googleapiclient.errors.HttpError as http_error:
-            # If the volume is not found, the API will raise
-            # googleapiclient.errors.HttpError.
-            cb.log.warning(
-                "googleapiclient.errors.HttpError: {0}".format(http_error))
+        vol = self.provider.get_resource('disks', volume_id)
+        if vol:
```
chiniforooshan: nit: "return return GCEVolume(self.provider, vol) if vol else None" fits in 1 line, right?

Similarly for bucket, network, region, image, and type (not snapshot and instance) if the short form fits in 1 line.

cloudbridge/cloud/providers/gce/provider.py
baizhang: Done.

cloudbridge/cloud/providers/gce/services.py
baizhang: Done.

baizhang:
Thanks for your review, @chiniforooshan !

Hi @nuwang @afgane @mbookman, could you please take a look at this pull request? Thanks!

nuwang:
@baizhang @chiniforooshan This is really nice - makes the code more concise and it all looks great! Thanks for implementing/reviewing!

PR #108, opened 2017-11-25
Azure provider merge ready

nuwang:
Need to resolve following issues:
https://github.com/gvlproject/cloudbridge/issues/107
https://github.com/gvlproject/cloudbridge/issues/106


cloudbridge/cloud/providers/azure/resources.py
```diff
@@ -0,0 +1,1781 @@
...
+        except CloudError as cloudError:
+            log.exception(cloudError.message)
+            return False
+
+    def detach(self, force=False):
+        """
+        Detach this volume from an instance.
+        """
+        virtual_machine = None
+        for index, vm in enumerate(
```
nuwang: @afgane I found this quite by luck because the linter complained that the index variable was unused. However, this method has severe issues - it is enumerating through every single vm to detach a disk and secondly, it's creating a vm a few lines below! Looks like we'll need to do a careful inspection of all the code.

cloudbridge/cloud/providers/azure/resources.py
```diff
@@ -0,0 +1,1781 @@
...
+
+            vm.storage_profile.data_disks.append({
+                'lun': len(vm.storage_profile.data_disks),
+                'name': self.id,
+                'create_option': 'attach',
+                'managed_disk': {
+                    'id': self.resource_id
+                }
+            })
+            self._provider.azure_client \
```
nuwang: Ditto here. Creating a VM!

cloudbridge/cloud/providers/azure/resources.py
nuwang: Ok. Looks like a false alarm for the latter. It's just a misnamed method - it should be create_or_update_vm.

nuwang: We may need to parameterize the detach method with the instance, because Azure doesn't seem to maintain a reference at the disk level: https://docs.microsoft.com/en-us/rest/api/compute/manageddisks/disks/disks-get

PR #114, opened 2018-02-15
GCE merge to master

nuwang:


cloudbridge/cloud/base/resources.py
```diff
@@ -782,6 +782,10 @@ def to_json(self):
         js['firewall'] = self.firewall.id
         return js
 
+    @property
+    def parent(self):
```
nuwang: @baizhang @chiniforooshan @mbookman Should the parent property be moved to the GCEFirewall class? It doesn't seem to apply to other providers?

setup.py
```diff
@@ -21,7 +21,8 @@
 REQS_BASE = [
     'bunch>=1.0.1',
     'six>=1.10.0',
-    'retrying>=1.3.3'
+    'retrying>=1.3.3',
+    'oslo.i18n>=3.15.3'
```
nuwang: @baizhang @chiniforooshan @mbookman Just wondering what this is currently used for?

setup.py
```diff
@@ -42,12 +44,15 @@
...
-REQS_FULL = REQS_BASE + REQS_AWS + REQS_AZURE + REQS_OPENSTACK
+REQS_FULL = REQS_BASE + REQS_AWS + REQS_AZURE + REQS_GCP + REQS_OPENSTACK
 # httpretty is required with/for moto 1.0.0 or AWS tests fail
 REQS_DEV = ([
     'tox>=2.1.1',
     'nose',
     # 'moto>=1.1.11',  # until https://github.com/spulec/moto/issues/1396
+    'docutils>=0.14',
+    'imagesize>=0.7.1',
+    'jinja2>=2.9.6',
```
nuwang: @baizhang @chiniforooshan @mbookman Also these - perhaps a comment indicating what they are used for?

test/helpers/__init__.py
```diff
@@ -142,7 +149,7 @@ def get_test_gateway(provider, name):
 
     This includes creating a network for the gateway, which is also returned.
     """
-    net_name = 'cb_testgwnet-{0}'.format(get_uuid())
+    net_name = 'cb-testgwnet-{0}'.format(get_uuid())
```
nuwang: @baizhang @chiniforooshan @mbookman Have we got the regex for name validation wrong? I thought gce supported underscores? If not, we may need to update the regex to reflect this.

test/test_network_service.py
```diff
@@ -43,15 +44,14 @@ def test_network_properties(self):
             sit.check_repr(self, net)
 
             self.assertIn(
-                net.cidr_block, ['', '10.0.0.0/16'],
+                net.cidr_block, ['', '10.0.0.0/16', '10.128.0.0/9'],
```
nuwang: Can you clarify why the second CIDR block is needed? It doesn't match the one we specified when creating the network?

test/test_network_service.py
```diff
@@ -43,15 +44,14 @@ def test_network_properties(self):
...
                 "Network CIDR %s does not contain the expected value."
                 % net.cidr_block)
 
-            cidr = '10.0.1.0/24'
-            sn = net.create_subnet(
-                name=subnet_name, cidr_block=cidr,
-                zone=helpers.get_provider_test_data(self.provider,
-                                                    'placement'))
+            cidr = '{0}/24'.format(net.cidr_block.split('/')[0] or '10.0.0.0')
```
nuwang: Also, this is a bit unclear.

test/helpers/__init__.py
afgane: GCE does not allow underscores but from our discussions, iirc, GCE provider was (for the resource name) going to replace the additional symbols allowed within. There's more about this in this comment: https://github.com/gvlproject/cloudbridge/issues/55#issuecomment-322810917

mbookman: GCE allows underscores and dashes in labels, but not in resources (instances, volumes, networks).

I think at the time we discussed https://github.com/gvlproject/cloudbridge/issues/55#issuecomment-322810917, we focused on instance names and agreed to:

- Will not allow upper case letters and will make the restrictions visible to the user.
- Will label field as "host name" or some such to make the restrictions more expected for the user.

The thought was that people generally understand host name restrictions and so pushing this restriction all the way up would be fine.

Would it make sense for each provider to have a validation routine for resource names?

chiniforooshan:
Thank you Nuwan! Let me send a few more PRs and then merge the gce branch into master. I'll let you know when we are ready.

test/helpers/__init__.py
afgane: This is a good point and it makes a case for not allowing underscores at all.

mbookman:
Hi @nuwang. What are the next steps to move this PR forward to get the gce branch merged to master?

nuwang:
@mbookman We think things are looking pretty fantastic, thanks to the wonderful and patient effort that the three of you have put into this. I think the outstanding work is:
1. Resolve conflicts
2. Review PR as a whole, mostly just the required interface changes and how global test changes could affect the existing code. As long as the tests pass, we should be ok.
3. Make sure all tests pass after merge to master.
4. Make sure docs are up-to-date with instructions on using the GCE provider

That's with regard to this PR itself. Afterwards, we will have to test the code against cloudlaunch and cloudman, and make the required changes to their respective UIs so that GCE credentials can be accepted etc. There's also the cloudbridge paper. We'll draft an initial version and send it along for review/modification.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -68,9 +69,76 @@ def vm_firewalls(self):
 
 class GCEKeyPairService(BaseKeyPairService):
 
+    GCEKeyInfo = namedtuple('GCEKeyInfo', 'format public_key email')
+
     def __init__(self, provider):
         super(GCEKeyPairService, self).__init__(provider)
 
+    def _iter_gce_key_pairs(self, provider):
```
afgane: What's the difference between this method and `_iter_gce_ssh_keys` below?

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -68,9 +69,76 @@ def vm_firewalls(self):
...
+            entry = ""
+            for gce_kp in gce_kp_list:
+                entry = entry + u"{0} {1} {2}\n".format(gce_kp.format,
+                                                        gce_kp.public_key,
+                                                        gce_kp.email)
+            sshkey_entry["value"] = entry.rstrip()
+
+        helpers.gce_metadata_save_op(provider, _process_kps_from_metadata)
+
+    def gce_kp_to_id(self, gce_kp):
```
afgane: This can yield multiple same IDs (when multiple same public keys are uploaded), which is against our definition of an ID as uniquely identifiable. Perhaps the ID here should be a combination of the `name` and `public_key` properties?

cloudbridge/cloud/providers/openstack/resources.py
```diff
@@ -1212,18 +1212,21 @@ def __init__(self, provider, vm_firewall):
...
         """
-        OpenStack does not associate a SG with a network so default to None.
+        OpenStack does not associate a fw with a network so extract from desc.
 
-        :return: Always return ``None``.
+        :return: The network ID supplied when this firewall was created or
+                 `None` if ID cannot be identified.
         """
         # Best way would be to use regex, but using this hacky way to avoid
         # importing the re package
```
afgane: Why not import `re`? Below split does not always work as per FIXME.

PR #116, opened 2018-02-22
Fixing some bugs

chiniforooshan:
Please see commit messages for more details.

@nuwang @mbookman @baizhang

### DIFF
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -1927,7 +1927,7 @@ class GCSObject(BaseBucketObject):
 
     def iter_content(self):
         return io.BytesIO(self._provider
-                              .gcp_storage
+                              .gcs_storage
                               .objects()
                               .get_media(bucket=self._obj['bucket'],
                                          object=self.name)
@@ -1958,7 +1958,7 @@ class GCSObject(BaseBucketObject):
 
     def delete(self):
         (self._provider
-             .gcp_storage
+             .gcs_storage
              .objects()
              .delete(bucket=self._obj['bucket'], object=self.name)
              .execute())
@@ -1987,7 +1987,7 @@ class GCSBucketContainer(BaseBucketContainer):
         max_result = limit if limit is not None and limit < 500 else 500
         try:
             response = (self._provider
-                            .gcp_storage
+                            .gcs_storage
                             .objects()
                             .list(bucket=self.bucket.name,
                                   prefix=prefix if prefix else '',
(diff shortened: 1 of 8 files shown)


cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1670,7 +1670,7 @@ def description(self, value):
 
     @property
     def size(self):
-        return self._volume.get('sizeGb')
+        return int(self._volume.get('sizeGb'))
```
nuwang: @chiniforooshan Is this guaranteed to be non-null? We had a few problems with OpenStack or AWS where some of these properties return null values and that results in an exception...

test/test_security_service.py
```diff
@@ -97,7 +97,7 @@ def test_vm_firewall_properties(self):
 
     @helpers.skipIfNoService(['security.vm_firewalls'])
     def test_crud_vm_firewall_rules(self):
-        name = 'cb-crudfw_rules-{0}'.format(helpers.get_uuid())
+        name = 'cb-crudfw-rules-{0}'.format(helpers.get_uuid())
```
nuwang: Just a note: May need a corresponding change in validation regex here: https://github.com/gvlproject/cloudbridge/blob/eb20e3d88f2893a6e281ceead7ab8fd3527be6af/cloudbridge/cloud/base/resources.py#L69

Or more appropriately, the GCE resources that do not support dashes can selectively override the CB_NAME_PATTERN property to implement the appropriate regex?

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: I've not seen a case where this is null. If it's OK with you, I prefer it breaks if it's null so that we see it, try to understand when it happens, and fix it properly.

test/test_security_service.py
chiniforooshan: Bai is going to make GCE resources override the pattern in his next PR.

test/test_compute_service.py
```diff
@@ -272,14 +272,18 @@ def cleanup_snap(snap):
                     "vm_type")
                 vm_type = self.provider.compute.vm_types.find(
                     name=vm_type_name)[0]
-                for _ in range(vm_type.num_ephemeral_disks):
+                # Some providers, e.g. GCP, has a limit on total number of
+                # attached disks; it does not matter how many of them are
+                # ephemeral or persistent. So, wee keep in mind that we have
+                # attached 4 disks already, and add ephemeral disks accordingly
+                # to not exceed the limit.
+                for _ in range(vm_type.num_ephemeral_disks - 4):
```
nuwang: Something seems wrong here? For example, if num_ephemeral_disks were 50, subtracting 4 would mean we'd still be over the limit? It looks like we need to document that there's a disk limit that applies across the whole of cloudbridge, and maybe enforce that?

test/test_security_service.py
nuwang: Sounds good.

PR #118, opened 2018-03-27
Use ipaddress instead of unmaintained py2-ipaddress 

nsoranzo:
Also build universal wheels.

Both will help the Galaxy under Python3 project.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,3 +14,5 @@ match=^[Tt]est
 # Don't capture stdout - print immediately
 nocapture=1
 
+[bdist_wheel]
+universal = 1
--- a/setup.py
+++ b/setup.py
@@ -63,8 +63,7 @@ setup(
     url='http://cloudbridge.cloudve.org/',
     install_requires=REQS_FULL,
     extras_require={
-        ':python_version=="2.7"': ['py2-ipaddress'],
-        ':python_version=="3"': ['py2-ipaddress'],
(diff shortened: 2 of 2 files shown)


nuwang:
@nsoranzo Thanks, this looks great! I guess the newer requirement syntax doesn't work on pypy yet, but we can fix that when it eventually does. I think the build failures on openstack and azure are unrelated, because we've run into a quota issue on them.

nsoranzo:
Everything works on PyPI, this is how I build and upload my Python packages:
```
$ python setup.py sdist
$ python setup.py bdist_wheel
$ twine upload dist/*
```

You just need recent versions of `pip`, `setuptools`, `twine` and `wheel`.

nuwang:
I was thinking about PyPy, as in the jit compiler - I saw that an earlier commit had a newer requirements syntax, which I assume you removed it because of PyPy?

nsoranzo:
Ah, sorry, I misread! The problem is not PyPy (that I know of), but old `setuptools`: the `extras_require` syntax is understood by setuptools >=18, while the PEP 508 syntax I initially used is understood only by setuptools >=36.2 . See https://hynek.me/articles/conditional-python-dependencies/ for more details.

It is also possible to support setuptools<18, if needed, with an ugly workaround as in https://gitlab.com/pycqa/flake8/blob/master/setup.py .

nuwang:
That makes sense - thanks for the details. PEP508 looks really interesting!

PR #121, opened 2018-04-28
Fix issues for GCE when running CloudBlockStoreServiceTestCase.

baizhang:

This pull request fixes a few issues for GCE when running CloudBlockStoreServiceTestCase. Specifically,
1. Make the delete method for GCEVolume and GCESnapshot synchronous (waiting for the operation to complete before returning).
2. Add resource description handling for GCEVolume and GCESnapshot
3. Check the validity of the name for name setter method.
4. Disable the renaming checks, since GCE does not support resource renaming after it is created.
5. Add cache_discovery=False to suppress error "ImportError: file_cache is unavailable when using oauth2client >= 4.0.0" ([link](https://github.com/google/google-api-python-client/issues/299)).

Thank you!

### DIFF
--- a/cloudbridge/cloud/providers/gce/resources.py
+++ b/cloudbridge/cloud/providers/gce/resources.py
@@ -758,6 +758,7 @@ class GCEInstance(BaseInstance):
         """
         Set the instance name.
         """
+        GCEInstance.assert_valid_resource_name(name)
         # In GCE, the name of the instance is provided by the client when
         # initially creating the resource. The name cannot be changed after
         # the instance is created.
@@ -1639,12 +1640,22 @@ class GCEVolume(BaseVolume):
         """
         return self._volume.get('name')
 
+    @name.setter
+    # pylint:disable=arguments-differ
+    def name(self, value):
+        GCEVolume.assert_valid_resource_name(value)
+        # In GCE, the name of the volume is provided by the client when
+        # initially creating the resource. The name cannot be changed after
+        # the volume is created.
+        cb.log.warning("Setting volume name after it is created is not "
+                       "supported by this provider.")
+
     @property
     def description(self):
         labels = self._volume.get('labels')
-        if not labels or 'description' not in labels:
-            return ''
-        return labels.get('description', '')
+        if labels and 'description' in labels:
+            return labels.get('description', '')
+        return self._volume.get('description', '')
 
     @description.setter
     def description(self, value):
@@ -1774,13 +1785,15 @@ class GCEVolume(BaseVolume):
         """
         Delete this volume.
         """
-        (self._provider
-             .gce_compute
-             .disks()
-             .delete(project=self._provider.project_name,
-                     zone=self._provider.default_zone,
-                     disk=self.name)
-             .execute())
+        response = (self._provider
+                    .gce_compute
+                    .disks()
+                    .delete(project=self._provider.project_name,
+                            zone=self._provider.default_zone,
+                            disk=self.name)
+                    .execute())
+        self._provider.wait_for_operation(
+            response, zone=self._provider.default_zone)
 
     @property
     def state(self):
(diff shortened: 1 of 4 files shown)


cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1639,12 +1640,22 @@ def name(self):
...
         return self._volume.get('name')
 
+    @name.setter
+    # pylint:disable=arguments-differ
+    def name(self, value):
+        GCEVolume.assert_valid_resource_name(value)
+        # In GCE, the name of the volume is provided by the client when
+        # initially creating the resource. The name cannot be changed after
+        # the volume is created.
+        cb.log.warning("Setting volume name after it is created is not "
```
nuwang: @baizhang Can we use a label/tag to track the name instead? In cloudbridge, the name is just a user provided friendly name, and is therefore, changeable. So we could do something similar to what's done for description below? (unless it's super inefficient, and requires a new http request)

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1774,13 +1785,15 @@ def delete(self):
...
-             .execute())
+        response = (self._provider
+                    .gce_compute
+                    .disks()
+                    .delete(project=self._provider.project_name,
+                            zone=self._provider.default_zone,
+                            disk=self.name)
+                    .execute())
+        self._provider.wait_for_operation(
+            response, zone=self._provider.default_zone)
```
nuwang: @baizhang Is there a specific reason to block till the delete is complete?? I think ideally, the delete call shouldn't block. I believe the tests handle this already, and wait for the delete to complete.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1818,16 +1831,46 @@ def id(self):
...
+    @name.setter
+    # pylint:disable=arguments-differ
+    def name(self, value):
+        GCESnapshot.assert_valid_resource_name(value)
+        # In GCE, the name of the snapshot is provided by the client when
+        # initially creating the resource. The name cannot be changed after
+        # the snapshot is created.
+        cb.log.warning("Setting snapshot name after it is created is not "
+                       "supported by this provider.")
+
```
nuwang: Same as for Volume name. The name should ideally be editable. In GCE, I guess the 'name' maps more to cloudbridge's id.

cloudbridge/cloud/providers/gce/resources.py
mbookman: Hi @nuwang -

In https://github.com/gvlproject/cloudbridge/pull/44, where we landed was:

> @nuwang - your proposal seems reasonable to me:
> 
> Use the disk name field for name (with the specified character and inability to rename restrictions)
> Use labels for the disk description field. GCE implementation will convert spaces to underscores
 
Please review the discussion. Do the premises of the discussion in that PR still hold or do we need to revisit?

Thanks,

-Matt

nuwang: @mbookman Thanks for pointing to that. Going through that discussion through, I can't find why exactly we decided to go with the GCE name for the 'name' field, instead of using a label, but I think there was some important reason? As I recall, @afgane raised an issue - something to do with potential mismatches between the original name given at creation time and a later rename perhaps? It would be good to record what it is for future reference.

mbookman: And - not all Cloud resource support labels. So we could potentially end up with a hodge podge of resources which did support renames vs. those that did not.

nuwang: @mbookman Sorry, I had missed this comment. Right, that was the issue - thanks for penning it down. I guess we still haven't figured out a way to make this consistent across all providers, since, as it stands now, a rename will fail silently for GCE only. Let's leave this as is for now, since it's not a major issue, and hopefully, figure out a way to make things uniform across all providers if and when needed?

cloudbridge/cloud/providers/gce/resources.py
nuwang: This is the only issue that remains - should we just go ahead and merge it and we can sort this issue out later since it isn't a blocker?

baizhang: Thanks for your review, @nuwang !

I agree. Let's go ahead and merge this one first. And I will work on a non-blocking delete method in a separate pull request.

PR #127, opened 2018-07-14
Azure gallery images for merge

nuwang:
@almahmoud I've cleaned up the commit history a bit using "git rebase -i" and also made two additional commits (https://github.com/CloudVE/cloudbridge/commit/3af813fa1a26e212973b82b1740ba79770a21c98 and  https://github.com/CloudVE/cloudbridge/pull/127/commits/0057fbc36f00ca18bc64df4fda69ed251c7d11d8) which does some minor refactoring. Let me know whether it all looks good and we can merge.


cloudbridge/cloud/providers/azure/azure_client.py
```diff
@@ -585,13 +587,52 @@ def delete_subnet(self, subnet_id):
...
+                timeout = self._config.get("default_wait_timeout")
+                interval = self._config.get("default_wait_interval")
+
+                assert timeout >= 0
+                assert interval >= 0
+                assert timeout >= interval
+
+                end_time = time.time() + timeout
+
+                while time.time() < end_time:
```
nuwang: Is it possible to use the tenacity package here? as in this commit: https://github.com/CloudVE/cloudbridge/pull/127/commits/f1ab169b199d6ce32ea22ca83b91cdea86a2a781

cloudbridge/cloud/providers/azure/azure_client.py
```diff
@@ -585,13 +587,52 @@ def delete_subnet(self, subnet_id):
...
+                .subnets.delete(
+                    self.resource_group,
+                    network_name,
+                    subnet_name
+                )
+            result_delete.wait()
+        except CloudError as cloud_error:
+            log.exception(cloud_error.message)
+            print(cloud_error.code)
+            if cloud_error.code != "InUseSubnetCannotBeDeleted":
```
nuwang: Why does this error get triggered? In principle, our tests should wait for the previous object to be deleted before attempting to delete the subnet (using the resource.wait_for method), so the subnet should not be in use any longer?

PR #130, opened 2018-07-24
GCP: Fixing tests and applying API changes

chiniforooshan:


chiniforooshan:
Hi Nuwan, Enis,

After these changes ALL tests pass when I use service account to authenticate. If I use application default credentials, only one test (test_generate_url) fails and it's because to generate a signed URL the private key is needed.

I'm not sure why the continuous integration tests fail on the PR. Could you please take a look?

Here is how I run tests:

source .vnv/bin/activate
cd cloudbridge
python setup.py install
pip install -U -e .[dev]
export CB_TEST_PROVIDER=gce
export GCE_SERVICE_CREDS_FILE=<path-to-creds-file>
tox -e py27

Thanks!
@nuwang @afgane @mbookman @baizhang

cloudbridge/cloud/base/resources.py
```diff
@@ -356,12 +356,16 @@ class BasePageableObjectMixin(PageableObjectMixin):
...
+            yield result
+
+    def iter(self, **kwargs):
+        result_list = self.list(**kwargs)
         if result_list.supports_server_paging:
             for result in result_list:
                 yield result
             while result_list.is_truncated:
-                result_list = self.list(marker=result_list.marker)
+                result_list = self.list(marker=result_list.marker, **kwargs)
```
nuwang: There's a bit of an issue here - at an interface level, the list method does not accept keyword args, it only accepts limit and marker. The list method does not support filtering of any kind. In contrast, the find() method accepts kwargs (and probably ought to have been renamed to filter instead). We think that perhaps the thing to do here is to move the iter() implementation into the GCE base class instead, and call self.find() instead of self.list(). At some future point, we can revisit the list method and consider whether to change the interface contract, but right now, it would require sweeping interface changes.

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -927,7 +966,9 @@ def key_pair_name(self):
         """
         Get the name of the key pair associated with this instance.
         """
-        return self._provider.security.key_pairs.name
+        for kp in self._provider.security.key_pairs:
```
nuwang: should this be return next(self._provider.security.key_pairs) with a try/catch?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -1483,10 +1521,14 @@ def attach_subnet(self, subnet):
...
-        cb.log.warning('GCE routers should be attached at creation time')
+        cb.log.warning('Google Cloud Routers automatically learn new subnets '
+                       'in your VPC network and announces them to your '
+                       'on-premises network')
 
     def detach_subnet(self, network_id):
-        cb.log.warning('GCE routers are always attached')
+        cb.log.warning('Cannot detach from subnet. Google Cloud Routers '
+                       'automatically learn new subnets in your VPC network '
+                       'and announces them to your on-premises network')
```
nuwang: Does this mean that there's no way to isolate two subnets in GCE? All subnets will automatically start communicating with each the moment a router is present?

cloudbridge/cloud/providers/gce/resources.py
```diff
@@ -2095,6 +2152,7 @@ def delete(self, delete_contents=False):
              .buckets()
              .delete(bucket=self.name)
              .execute())
+        time.sleep(2)
```
nuwang: Arbitrary timeout?

chiniforooshan:
Thanks for the comments!

test/test_security_service.py
```diff
@@ -204,9 +204,9 @@ def test_vm_firewall_group_rule(self):
                 r.delete()
             fw = self.provider.security.vm_firewalls.get(fw.id)  # update
             self.assertTrue(
-                len(list(fw.rules)) == 0,
+                fw is None or len(list(fw.rules)) == 0,
```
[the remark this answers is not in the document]
chiniforooshan: It is already the case that we create a dummy rule when we create a firewall, which simplified a lot of other tests. But, this particular test is explicitly deleting all rules, including the dummy rule that was created (lines 203 and 204). That's why vm_firewalls.get may return None.

test/test_network_service.py
```diff
@@ -71,7 +70,7 @@ def test_network_properties(self):
                     "Network ID %s should be specified in the subnet's network"
                     " id %s." % (net.id, sn.network_id))
 
-                self.assertEqual(
+                self.assertTrue(
```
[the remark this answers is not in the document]
chiniforooshan: Done. Generally, if there is already a subnetwork with overlapping CIDR block, then create_subnet does not create a new subnetwork and returns the existing one. But, this should not happen in this test.

test/test_block_store_service.py
```diff
@@ -116,7 +116,7 @@ def test_volume_properties(self):
                 self.assertEqual(test_vol.attachments.instance_id,
                                  test_instance.id)
                 if (self.provider.PROVIDER_ID != 'azure' and
-                    self.provider.PROVIDER_ID != 'gce'):
+                        self.provider.PROVIDER_ID != 'gce'):
```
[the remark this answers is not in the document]
chiniforooshan: You can programmatically call vol.attachments.device; however, attached devices on GCE are mounted like /dev/disk/by-id/google-*. It will not be /dev/sda2, as the test expects.

cloudbridge/cloud/providers/gce/services.py
```diff
@@ -1225,6 +1252,7 @@ def create(self, name, location=None):
                             .execute())
             if 'error' in response:
                 return None
+            time.sleep(2)
```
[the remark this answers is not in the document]
chiniforooshan: This is not completely arbitrary. GCS has a rate limit of 1 operation per 2 seconds for bucket creation/deletaion: https://cloud.google.com/storage/quotas. So the 2s throttling is to avoid future failures.

mbookman:
Looks like the travis tests are failing because there are no credentials for travis to use.
I think @afgane or @nuwang need to set this up.

Instructions for this setup can be found here:

https://cloud.google.com/solutions/continuous-delivery-with-travis-ci#creating_the_credentials

test/test_block_store_service.py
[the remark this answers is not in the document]
mbookman: GCE doesn't allow you to specify an explicit path, though you can specify a device name.
When you specify the name, then you can assume an explicit path `/dev/disk/by-id-google-<name>` is created. Underneath this is a link to `/dev/sda<char>` where the characters start with `a` for the first disk (the boot disk) and `b`, `c`, etc. for subsequent disks.
See https://medium.com/@DazWilkin/compute-engine-identifying-your-devices-aeae6c01a4d7.

However - Is the ability to specify a specific device path important for the volume API?
If so, then I think it would be preferred for the create() API to accept a device name and then have an implementation-specific device path returned.
That said - I'm not familiar with the Azure API and how it would fit here.

test/test_security_service.py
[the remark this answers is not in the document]
nuwang: Right, in the other cloud providers, a firewall with no rules will continue to exist. Can we change GCE behaviour to match so that deleting all the rules will leave the dummy rule intact, so that the firewall will continue to exist? This will make the behaviour consistent with other providers. Calling firewall.delete() can delete the remaining dummy rule, permanently removing the firewall from existence.

cloudbridge/cloud/providers/gce/services.py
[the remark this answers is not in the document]
nuwang: Ok, that makes sense. I think it would be great if we can add a comment in the code detailing that rationale, just so we don't trip up on its necessity in future.

test/test_block_store_service.py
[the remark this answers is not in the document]
nuwang: @chiniforooshan @mbookman That makes sense. It looks like we should change the vol.attach() interface to not accept the device id as a parameter. Is there a suitable way we could generalize this test but still verify that attachments.device remains a reasonable value, and not null/empty or invalid. I suppose we could just have a regex testing for a slash followed by a slash?

test/test_network_service.py
```diff
@@ -170,11 +169,11 @@ def _cleanup(net, subnet, router, gateway):
...
-            self.assertEqual(
-                router.state, RouterState.DETACHED,
-                "Router {0} state {1} should be {2}.".format(
-                    router.id, router.state, RouterState.DETACHED))
-
+#            self.assertEqual(
+#                router.state, RouterState.DETACHED,
+#                "Router {0} state {1} should be {2}.".format(
+#                    router.id, router.state, RouterState.DETACHED))
+#
```
nuwang: Should this test be reinstated? Or is the test invalid?

cloudbridge/cloud/base/resources.py
nuwang: @chiniforooshan What are your thoughts on this?

test/test_security_service.py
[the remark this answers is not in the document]
chiniforooshan: Let me try to understand the suggestion: are you saying that calling r.delete() on the dummy rule should not delete it? If so, then this test should assert that len(list(fw.rules)) == 1 not 0, because the dummy rule will still exist. Maybe you think fw.rules should not return the dummy rule? Then, that contradicts the test a few lines about this test that asserts that len(list(fw.rules)) > 0 right after creation of the firewall before explicitly adding any rules, which implies that fw.rules should indeed return the dummy rules?

It appears to me that fw is None or len(list(fw.rules)) == 0 is the most consistant behaviour we can get :)

cloudbridge/cloud/providers/gce/services.py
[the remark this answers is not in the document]
chiniforooshan: Done.

cloudbridge/cloud/base/resources.py
chiniforooshan: The list() method does support filtering, even at an interface level, for some services, e.g. SubnetService. So, IMHO, this is the right place for iter(). WDYT?

Also, the find() methods' return types seem to be inconsistent at the interface level; sometimes they return lists, sometimes they return a single object.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: As far as I understand it, GCP routers are for controlling the traffic between your GCP network and your on-premises networks, not for controlling traffic among subnets in your GCP network.

From the interface itself, it's not clear to me what is intended by router.{attach/detach}_subnet methods.

If you want to isolate some subnets from on-premises clients, we can use the custom mode of the GCP routers and advertise specific subnets explicitly instead of relying on the automatic detection of subnets, which is the default behaviour (https://cloud.google.com/router/docs/how-to/advertising-subnets). Should I do this?

@mbookman for opinions and verifying what I said.

cloudbridge/cloud/providers/gce/resources.py
chiniforooshan: Added comments.

cloudbridge/cloud/base/resources.py
nuwang: Hmm.. you're right. That is a bit inconsistent. The list method should ideally have a fixed signature without filtering capabilities. The find method is for filtering and should return a list, never a single record. The plan was to eventually have the find method superseded by chainable filters, mimicking what django, boto3 etc. do. I'm ok with leaving this as is for now, this probably needs to be revised across the whole of cloudbridge.

test/test_block_store_service.py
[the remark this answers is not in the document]
nuwang: That's true, the ideal test would have been an ssh based test - but it was never done because it seemed like a lot of work for relatively little gain. I think your suggestion makes sense in the meantime, thanks.

test/test_security_service.py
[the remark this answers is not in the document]
nuwang: Yes, I was thinking that all find, list methods etc. would always transparently drop the dummy rule before returning values. You're right, the asserts len(list(fw.rules)) > 0 rule is wrong. It was put in place because aws and openstack both add some default rules when a firewall is created. I think the default is something like a "deny all" rule or something. If GCE does not do the same, then that test is invalid. As long as our dummy rule is effectively a no-op, I think we should be able to safely drop it?

cloudbridge/cloud/providers/gce/resources.py
nuwang: The intent of router.attach_subnet is to allow routing between subnets as well as to route external gateway traffic to the subnet. We definitely haven't considered routing between on-premise and VPC yet, so this is more about subnet to subnet comms.

test/test_security_service.py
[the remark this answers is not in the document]
chiniforooshan: Thanks. I deleted the assertion of len(list(fw.rules)) > 0 after a firewall creation and changed the dummy rule behaviour accordingly.

cloudbridge/cloud/providers/gce/resources.py
mbookman: It looks like instead of creating a Google Cloud Router, the GCE implementation here should be updating the routing table that is associated with the VPC:

https://cloud.google.com/vpc/docs/routes

I *think* this is similar to the AWS implementation, if I'm reading it correctly.

mbookman: @nuwang : can you point to an explanation of the use case for the routing between subnets and the internet and providing end user control? Want to make sure that the implementation is matches the expected semantics.

Also - I believe this is the last issue currently for this PR. It would be great if we could come up with a solution to get this PR merged.

nuwang: @mbookman Sorry, I misspoke - the main requirement behind the router was subnet<->internet comms, not subnet<->subnet comms. We are currently not using any kind of subnet<->subnet comms and with Kubernetes, don't foresee an immediate need for it. Nevertheless, conceptually, if all subnets connected to the same router can route between each other, that would be desirable, but not critical.

nuwang:
@chiniforooshan This all looks good, thanks for adding the dummy rule support. Should we go ahead and merge?

chiniforooshan:
Yeah, I think this is ready to be merged. Thank you!

PR #136, opened 2018-08-07
Token-based access for Azure resources

VJalili:
If user provides an optional `access_token`, an instance of `BlockBlobService` is created using this token. The `ServicePrincipalCredentials` automatically creates an access token (`self._credentials.token`), and it follows a very similar logic to `CloudAuthz` to create this token: 

https://github.com/Azure/msrestazure-for-python/blob/2ab56d8f24042fe4d5511ba05550ac004e685d3d/msrestazure/azure_active_directory.py#L362-L368

https://github.com/galaxyproject/cloudauthz/blob/2ee68ab187661136e3f0640e1bc1225b201e6441/cloudauthz/providers/azure.py#L20-L26

However, `ServicePrincipalCredentials` makes some changes to the generated token (see [these lines](https://github.com/Azure/msrestazure-for-python/blob/2ab56d8f24042fe4d5511ba05550ac004e685d3d/msrestazure/azure_active_directory.py#L171-L174)), and CloudAuthz does not change generated token. Hence, I have not been able to use the token generate by `ServicePrincipalCredentials` to read/write to Azure blob storage. Therefore, in this PR I'm relying on the token user passes to cloudbridge. 

An example of a payload with `access_token` is as the following:
```python
config = {
    'azure_subscription_id': "...", 
    'azure_client_id': "...",
    'azure_secret': "...",
    'azure_tenant': "...",
    'azure_storage_account': '...',
    'azure_resource_group': '...',
    'azure_access_token': '...'}
```

### DIFF
--- a/cloudbridge/cloud/providers/azure/azure_client.py
+++ b/cloudbridge/cloud/providers/azure/azure_client.py
@@ -158,7 +158,7 @@ class AzureClient(object):
             tenant=config.get('azure_tenant')
         )
 
-        self._access_token = config.get('access_token')
+        self._access_token = config.get('azure_access_token')
         self._resource_client = None
         self._storage_client = None
         self._network_management_client = None
--- a/cloudbridge/cloud/providers/azure/azure_client.py
+++ b/cloudbridge/cloud/providers/azure/azure_client.py
@@ -158,7 +158,7 @@ class AzureClient(object):
             tenant=config.get('azure_tenant')
         )
 
-        self._resource_client = None
+        self._access_token = config.get('access_token')
         self._resource_client = None
         self._storage_client = None
         self._network_management_client = None
--- a/cloudbridge/cloud/providers/azure/azure_client.py
+++ b/cloudbridge/cloud/providers/azure/azure_client.py
@@ -255,8 +255,8 @@ class AzureClient(object):
                     token_credential=token_credential)
             else:
                 self._block_blob_service = BlockBlobService(
-                    self.storage_account,
-                    self.access_key_result.keys[0].value)
+                    account_name=self.storage_account,
+                    account_key=self.access_key_result.keys[0].value)
         return self._block_blob_service
 
     @property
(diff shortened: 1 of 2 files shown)


cloudbridge/cloud/providers/azure/azure_client.py
```diff
@@ -157,6 +158,7 @@ def __init__(self, config):
             tenant=config.get('azure_tenant')
         )
 
+        self._resource_client = None
```
nuwang: Redundant line?

cloudbridge/cloud/providers/azure/azure_client.py
```diff
@@ -246,9 +248,15 @@ def network_management_client(self):
...
-                self.access_key_result.keys[0].value)
+            if self._access_token:
+                token_credential = TokenCredential(self._access_token)
+                self._block_blob_service = BlockBlobService(
+                    account_name=self.storage_account,
+                    token_credential=token_credential)
+            else:
+                self._block_blob_service = BlockBlobService(
+                    self.storage_account,
+                    self.access_key_result.keys[0].value)
```
nuwang: it would be great if we could use named parameters here too, so that it's clear what the difference between the two variants.

cloudbridge/cloud/providers/azure/azure_client.py
VJalili: This is was an issue in my branch, it if fixed now.

cloudbridge/cloud/providers/azure/azure_client.py
VJalili: Agree, added.

VJalili:
Note: regardless of the access token, you still need to provide all the required credentials as they are required to instantiate `ServicePrincipalCredentials`: 

https://github.com/CloudVE/cloudbridge/blob/09e7e2497bbdfe82f3db4047c6768a4ac24a7e54/cloudbridge/cloud/providers/azure/azure_client.py#L154-L158  

However, when the `access_token` is provided, cloudbridge uses that token to authorize access to Azure blob storage:

https://github.com/VJalili/cloudbridge/blob/e1e31da4e0ed3b74f64166d2dad4f1f21db98011/cloudbridge/cloud/providers/azure/azure_client.py#L251-L259

PR #139, opened 2018-08-15
Change Resource Group and Storage Account Handling

almahmoud:
Moved storage account creation to azure client so it is not created with the provider object.
Wrapped known exceptions for resource group and storage account creation with more detailed messages.
(Also includes #136 merge)

### DIFF
--- a/cloudbridge/cloud/providers/azure/azure_client.py
+++ b/cloudbridge/cloud/providers/azure/azure_client.py
@@ -12,13 +12,14 @@ from azure.mgmt.resource.subscriptions import SubscriptionClient
 from azure.mgmt.storage import StorageManagementClient
 from azure.storage.blob import BlobPermissions
 from azure.storage.blob import BlockBlobService
-
-from cloudbridge.cloud.interfaces.exceptions import WaitStateException
+from azure.storage.common import TokenCredential
 
 from msrestazure.azure_exceptions import CloudError
 
 import tenacity
 
+from cloudbridge.cloud.interfaces.exceptions import \
+    InvalidNameException, ProviderConnectionException, WaitStateException
 from . import helpers as azure_helpers
 
 log = logging.getLogger(__name__)
@@ -157,6 +158,7 @@ class AzureClient(object):
             tenant=config.get('azure_tenant')
         )
 
+        self._access_token = config.get('azure_access_token')
         self._resource_client = None
         self._storage_client = None
         self._network_management_client = None
@@ -165,6 +167,7 @@ class AzureClient(object):
         self._access_key_result = None
         self._block_blob_service = None
         self._table_service = None
+        self._storage_account = None
 
         log.debug("azure subscription : %s", self.subscription_id)
 
@@ -245,14 +248,22 @@ class AzureClient(object):
 
     @property
     def blob_service(self):
+        self._get_or_create_storage_account()
         if not self._block_blob_service:
-            self._block_blob_service = BlockBlobService(
-                self.storage_account,
-                self.access_key_result.keys[0].value)
+            if self._access_token:
+                token_credential = TokenCredential(self._access_token)
+                self._block_blob_service = BlockBlobService(
+                    account_name=self.storage_account,
+                    token_credential=token_credential)
+            else:
+                self._block_blob_service = BlockBlobService(
+                    account_name=self.storage_account,
+                    account_key=self.access_key_result.keys[0].value)
         return self._block_blob_service
 
     @property
     def table_service(self):
+        self._get_or_create_storage_account()
         if not self._table_service:
             self._table_service = TableService(
                 self.storage_account,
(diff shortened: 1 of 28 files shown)


cloudbridge/cloud/providers/azure/provider.py
```diff
@@ -113,25 +115,29 @@ def _initialize(self):
...
-            }
-            self._azure_client.create_storage_account(self.storage_account,
-                                                      storage_account_params)
+
+        except CloudError as cloud_error:
+            if "not be found" in cloud_error.message:
+                resource_group_params = {'location': self.region_name}
+
+                try:
+                    self._azure_client.\
```
nuwang: Do we have a race condition identical to the get_or_create_storage situation here? Wrap with tenacity perhaps?

cloudbridge/cloud/providers/azure/azure_client.py
```diff
@@ -278,6 +290,39 @@ def create_storage_account(self, name, params):
...
+                    reraise=True)
+    def _get_or_create_storage_account(self):
+        if self._storage_account:
+            return self._storage_account
+        else:
+            try:
+                self._storage_account = \
+                    self.get_storage_account(self.storage_account)
+            except CloudError as cloud_error:
+                if "not found" in cloud_error.message:
```
nuwang: Is there something like an error code we could use here instead? Is the error message potentially localized, and therefore language dependent?

cloudbridge/cloud/providers/azure/azure_client.py
```diff
@@ -278,6 +290,39 @@ def create_storage_account(self, name, params):
...
+                        'sku': {
+                            'name': 'Standard_LRS'
+                        },
+                        'kind': 'storage',
+                        'location': self.region_name,
+                    }
+                    self._storage_account = \
+                        self.create_storage_account(self.storage_account,
+                                                    storage_account_params)
+                elif "is already taken" in cloud_error.message:
```
nuwang: Won't this exception be raised only when creating a storage account, not when retrieving one?

almahmoud: Yes, this part wasn't finished, sorry about that. I was just pushing to the new branch before bed to launch a travis test so I don't have to leave my laptop open or wait for the tests to run before going to sleep.

cloudbridge/cloud/providers/azure/provider.py
almahmoud: We haven't had that happen so far. Do you think we should add it preventively?

cloudbridge/cloud/providers/azure/azure_client.py
almahmoud: They have status codes which are 404 for not found and 403 for permission denied, and error codes which are ResourceNotFound/ResourceGroupNotFound and AuthorizationFailed. I'm wary of using the status codes since a 404 could mean that the resource group doesn't exist or that the storage account doesn't exist when trying to get a storage account within a resource group, but I think we could use ResourceNotFound/ResourceGroupNotFound since there shouldn't be any other resource than the one we're getting that could cause this in that particular spot.

cloudbridge/cloud/providers/azure/azure_client.py
```diff
@@ -286,6 +290,65 @@ def create_storage_account(self, name, params):
...
+                            mess = 'The following error was returned by ' \
+                                   'Azure:\n%s\n\nThis is likely because the' \
+                                   ' Role associated with the provided ' \
+                                   'credentials does not allow for Storage ' \
+                                   'Account creation.\nA Storage Account is ' \
+                                   'necessary in order to perform the ' \
+                                   'desired operation. You must either ' \
+                                   'provide an existing Storage Account name' \
+                                   ' as part of the configuration, or ' \
+                                   'elevate the associated Role.\nFor more ' \
```
nuwang: I would propose removing the advice to elevate the associated Role, since we don't really endorse that? We could just suggest that they should provide a SA name as part of the cloudbridge config?

nuwang: Since what we are trying to achieve with cloudbridge is to make it easy for someone to write code that works on multiple clouds, without having to do much testing etc., someone who is not that familiar with azure should be naturally guided towards the right approach, whatever that might be. My understanding was that the right approach here was to specify the storage account name. Is that not the case?

I think it's ok to leave the instructions on elevating privileges etc. as long as we point out what the "proper"/recommended approach is. Right now, I'm not really sure how role-wise, the creation of the storage account is separated from the usage of a storage account, i.e. who does what from a practical standpoint. Does this mean that some higher privilege user creates the storage account for you, and then, at launch time, a cloudlaunch user simply specifies the SA name? Does that mean we need a GUI element in cloudlaunch to specify the SA name?

afgane: > My understanding was that the right approach here was to specify the storage account name. Is that not the case?

There's not a right and wrong way here. For simplicity sake of creating a provider config, we suggest a user specifies only the 4 required fields and a SA will be reused or created. However, creating it requires _owner_ or _contributor_ level role. A user may or may not have that privilege.

> Does this mean that some higher privilege user creates the storage account for you, and then, at launch time, a cloudlaunch user simply specifies the SA name? 

Possibly. If a user doesn’t have the permissions, they'll have to supply a SA name; if they don't supply it, we/CloudBridge will try to create it and possibly fail with the above message indicating to the user they have two two options.

> Does that mean we need a GUI element in cloudlaunch to specify the SA name?

Yes, just a another (optional) text field in the provider credentials form.

In general, I feel an elaboration of all of this belongs in the docs, not a comment in the exception. There, we can have a more explanatory text with the recommended approach while this here is just an informative error message that ideally gives an unsuspecting user a path forward.

PR #146, opened 2018-09-07
Documentation

almahmoud:
Finalizing documentation for updated interface


afgane:
I think after all we agreed here that `get` would only get by id and not name, right? Can update this then to remove the changes to `resources.py` files.

almahmoud:
Opened a branch for doc updates and separated code changes

PR #147, opened 2018-09-10
Azure updates

almahmoud:
-Unlike other resources, images can be shared between resource groups, thus I changed the image get method to parse both resource group name and image name, in case the ID points to a different resource group.
-Subnet labels are stored in Network tags because subnets don't support tags themselves. In order to avoid reaching the 15 tag limit per resource, subnet labels now share the same tag, up to the character limit, after which they start using a new tag.

### DIFF
--- a/cloudbridge/cloud/providers/azure/resources.py
+++ b/cloudbridge/cloud/providers/azure/resources.py
@@ -1247,6 +1247,7 @@ class AzureSubnet(BaseSubnet):
 
     def delete(self):
         self._provider.azure_client.delete_subnet(self.id)
+        self._network._remove_sn_label(self._subnet.name)
 
     @property
     def state(self):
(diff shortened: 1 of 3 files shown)


cloudbridge/cloud/providers/azure/resources.py
```diff
@@ -955,6 +955,91 @@ def create_subnet(self, label, cidr_block, zone=None):
...
+
+    def _add_sn_label(self, sn_name, sn_label):
+        # Refresh net to make sure tags are updated
+        self.refresh()
+        self.wait_till_ready()
+        # If the subnet is already tagged, replace the current value in the
+        # same tag
+        curr_tag = self._existing_sn_tag(sn_name)
+        if curr_tag:
+            self._replace_sn_label(curr_tag, sn_name, sn_label)
```
nuwang: if two different operations start updating subnet labels on the same network together, we have a potential race condition and potential lost update issue when these values are concatenated. I think the safer option is to revert to having a unique tag per subnet, although that restricts the number of subnets per network to 14 minus any used tags. Alternatively, we just drop label support altogether for subnets. I'm leaning more heavily in favour of the latter.

PR #152, opened 2018-10-03
Azure patch

almahmoud:
Fixing the label bug by deleting subnet label (limit will be 14 labeled subnets per network)
Implementing server-side paging for Azure buckets

### DIFF
--- a/cloudbridge/cloud/providers/azure/provider.py
+++ b/cloudbridge/cloud/providers/azure/provider.py
@@ -81,7 +81,9 @@ class AzureCloudProvider(BaseCloudProvider):
         else:
             return default
 
-    @deprecated(deprecated_in='1.1', current_version=cloudbridge.__version__,
+    @deprecated(deprecated_in='1.1',
+                removed_in='2.0',
+                current_version=cloudbridge.__version__,
                 details='AZURE_VM_DEFAULT_USER_NAME was deprecated in favor '
                         'of AZURE_VM_DEFAULT_USERNAME')
     def __wrap_deprecated_username(self, username):
--- a/cloudbridge/cloud/providers/azure/provider.py
+++ b/cloudbridge/cloud/providers/azure/provider.py
@@ -54,8 +54,8 @@ class AzureCloudProvider(BaseCloudProvider):
                                str(self.resource_group)))[-6:]))
 
         self.vm_default_user_name = self._get_config_value(
-            'azure_vm_default_user_name', get_env(
-                'AZURE_VM_DEFAULT_USER_NAME', 'cbuser'))
+            'azure_vm_default_username', get_env(
+                'AZURE_VM_DEFAULT_USERNAME', 'cbuser'))
 
         self.public_key_storage_table_name = self._get_config_value(
             'azure_public_key_storage_table_name', get_env(
--- a/cloudbridge/cloud/providers/azure/provider.py
+++ b/cloudbridge/cloud/providers/azure/provider.py
@@ -1,10 +1,13 @@
 import logging
 import uuid
 
+from deprecation import deprecated
+
 from msrestazure.azure_exceptions import CloudError
 
 import tenacity
 
+import cloudbridge
 from cloudbridge.cloud.base import BaseCloudProvider
 from cloudbridge.cloud.base.helpers import get_env
 from cloudbridge.cloud.interfaces.exceptions import ProviderConnectionException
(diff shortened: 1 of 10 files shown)


cloudbridge/cloud/providers/azure/services.py
```diff
@@ -1037,6 +1042,11 @@ def get_or_create_default(self, zone):
     def delete(self, subnet):
         subnet_id = subnet.id if isinstance(subnet, Subnet) else subnet
         self.provider.azure_client.delete_subnet(subnet_id)
+        network = subnet._network
+        az_network = network._network
+        az_network.tags.pop(subnet.tag_name)
+        self._provider.azure_client.update_network_tags(
+            az_network.id, az_network)
```
nuwang: @almahmoud Was this caught in a test? If not, can we write a test that would catch it?

almahmoud: It was caught when we hit the limit after running multiple tests. I can write one specifically for it, but it would be an Azure specific test, since there is no way of really checking for a label in OS or AWS cause the objects don't exist anymore.

cloudbridge/cloud/providers/azure/provider.py
```diff
@@ -54,8 +54,8 @@ def __init__(self, config):
                                str(self.resource_group)))[-6:]))
 
         self.vm_default_user_name = self._get_config_value(
-            'azure_vm_default_user_name', get_env(
-                'AZURE_VM_DEFAULT_USER_NAME', 'cbuser'))
+            'azure_vm_default_username', get_env(
```
afgane: Didn't we agree that this should accept both versions for backward compatibility, with the `azure_vm_default_user_name` being deprecated?

cloudbridge/cloud/providers/azure/services.py
afgane: Feels to me this could also use a comment as to why it's necessary to update the network tags.

nuwang: Yes, I guess there's no real way to test this in a cloud agnostic way, so perhaps we need not bother. Agree that a comment would be useful.

almahmoud: Added the comment in https://github.com/CloudVE/cloudbridge/pull/152/commits/24d5b1f0e7449422c03e975851232217d67881e4

cloudbridge/cloud/providers/azure/provider.py
almahmoud: Added in https://github.com/CloudVE/cloudbridge/pull/152/commits/742ca14de557f6f34c449b7fa2b5374a78f2bf2d

afgane:
Should we decide now at what version we'll remove the deprecated items?
The pieces for deprecation tracking seem like they're complicating the code so I'm in favor or having short deprecation cycles for simple things like argument name changes, particularly early in the life of the library.

cloudbridge/cloud/providers/openstack/services.py
```diff
@@ -210,14 +210,14 @@ def list(self, limit=None, marker=None):
         return ClientPagedResultList(self.provider, firewalls,
                                      limit=limit, marker=marker)
 
-    def create(self, label, description, network_id):
+    @cb_helpers.deprecated_alias(network_id='network')
```
nuwang: @almahmoud I think the original order is more intuitive when reading the code? 
```suggestion
    @cb_helpers.deprecated_alias(network='network_id')
```

nuwang:
> Should we decide now at what version we'll remove the deprecated items?
I think that sounds good. The deprecated decorator supported a removed_in parameter, so we should be able to specify which version we plan to completely remove it in. I think 1.2 or so should be fine, even though it's an interface change.

cloudbridge/cloud/providers/openstack/services.py
almahmoud: It didn't work properly with that order though since it was trying to convert network to network_id, and that parameter wasn't recognized. We could potentially switch it by flipping things in the helper function, but it seems like a lot of work for just aesthetics imo.

afgane: I agree it's not worth extra effort, especially given it'll be removed one day. Meanwhile, reading `@cb_helpers.deprecated_alias(network_id='network')` seems to me like the deprecated alias is `network_id` and it should be matched to `network`, no?

afgane:
Let's stick to semver then and just plan on removing it in 2.0.

Issue #153, opened 2018-10-03
Show instance types/vm sizes based on a selected region 


felix-wolfheimer:
This is based on issue #148. 

The list of instance types is currently independent of a selected region/location. However, for AWS and other providers the available instance types depend on the selected region. This is especially true for the "high-end" instance types like p3 in AWS. It would be great to have an API to find out which instance types are available in which region. Azure has an API for this which takes the "region" as argument and here's the code I use for Azure:

```
def GetInstanceTypesForRegion(self, region):
       
       from azure.mgmt.compute import ComputeManagementClient

       client = ComputeManagementClient(self._credentials, self._subscription_id)
       itypes = list()
       for elem in client.virtual_machine_sizes.list(region):
          # Unfortunately we need to rely on a weird logic to get the number of gpus for an instance.
          # GPUs are only available on NC, ND, and NV instances and there's a 6:1 ratio between CPUs and GPUs
          # for these instances.
          gpus = 0
          if (re.match(r'Standard_NC.*',elem.name) is not None) or (re.match(r'Standard_ND.*',elem.name) is not None) or (re.match(r'Standard_NV.*',elem.name) is not None):
             gpus = int(elem.number_of_cores/6)
          descr = '{},{},{},{:.1f}'.format(elem.name, elem.number_of_cores, gpus, elem.memory_in_mb)
          itypes.append(descr)
```

Here's the code I use for AWS. This code queries the pricing API and was the best way I could find to get some information about available instance types. However, this is far from optimal as it tends to not find all available instance types for some reason and the pricing API uses different names for the regions, such that I have to translate "EU (Frankfurt) -> eu-central-1" etc.  

```
def GetInstanceTypesForRegion(self, region):
      
       import boto3
       import demjson

       pretty_region_names = { 'ap-south-1'     : 'Asia Pacific (Mumbai)' ,
                                     'ap-northeast-1' : 'Asia Pacific (Tokyo)',
                                     'ap-northeast-2' : 'Asia Pacific (Seoul)',
                                     'ap-southeast-1' : 'Asia Pacific (Singapore)',
                                     'ap-southeast-2' : 'Asia Pacific (Sydney)',
                                     'ca-central-1'   : 'Canada (Central)',
                                     'eu-central-1'   : 'EU (Frankfurt)',
                                     'eu-west-1'      : 'EU (Ireland)',
                                     'eu-west-2'      : 'EU (London)',
                                     'eu-west-3'      : 'EU (Paris)',
[...]

PR #155, opened 2018-10-03
Filter out instance type by region.

nuwang:
Fixes: https://github.com/CloudVE/cloudbridge/issues/153

### DIFF
--- a/cloudbridge/cloud/providers/aws/services.py
+++ b/cloudbridge/cloud/providers/aws/services.py
@@ -616,13 +616,16 @@ class AWSVMTypeService(BaseVMTypeService):
         https://github.com/powdahound/ec2instances.info (in particular, this
         file: https://raw.githubusercontent.com/powdahound/ec2instances.info/
         master/www/instances.json).
-
-        TODO: Needs a caching function with timeout
         """
         r = requests.get(self.provider.config.get(
             "aws_instance_info_url",
             self.provider.AWS_INSTANCE_DATA_DEFAULT_URL))
-        return r.json()
+        # Some instances are only available in certain regions. Use pricing
+        # info to determine and filter out instance types that are not
+        # available in the current region
+        vm_types_list = r.json()
+        return [vm_type for vm_type in vm_types_list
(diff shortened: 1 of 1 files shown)


nuwang:
There's a small issue with this fix. Some instance types in our data source does not have pricing data for certain instance_types, for example the t3 series. As a result, they are always filtered out, although us-east-1 for example, supports t3.

This is the full set of instance types that got completely filtered out in us-east-1, although I think us-east-1 supports almost all instance types. If we can live without these types altogether, this would certainly reduce issues with unsupported instance types. Eventually, the data source may get updated to include price data for this.

```
{'f1.4xlarge',
 'r5.12xlarge',
 'r5.24xlarge',
 'r5.2xlarge',
 'r5.4xlarge',
 'r5.large',
 'r5.xlarge',
 'r5d.12xlarge',
 'r5d.24xlarge',
 'r5d.2xlarge',
 'r5d.4xlarge',
 'r5d.large',
 'r5d.xlarge',
 't3.2xlarge',
 't3.large',
 't3.medium',
 't3.micro',
 't3.nano',
 't3.small',
 't3.xlarge',
 'u-12tb1.metal',
 'u-6tb1.metal',
 'u-9tb1.metal',
 'z1d.12xlarge',
 'z1d.2xlarge',
 'z1d.3xlarge',
 'z1d.6xlarge',
 'z1d.large',
 'z1d.xlarge'}
```

afgane:
I feel t3 in particular is a popular instance type. For Galaxy, r5 is probably desirable as well so I don't really think we should be filtering those by default.

nuwang:
In that case, the only recourse seems to be to make a PR/patch against the data source for all the instance types we need, before merging this in. I think it would be good opportunity to contribute back to that project too.

nuwang:
Looks like it's already been worked on: https://github.com/powdahound/ec2instances.info/issues/360

nuwang:
The missing instances have has been fixed in: https://github.com/powdahound/ec2instances.info/pull/389, so I'll go ahead and merge this.

PR #171, opened 2019-02-03
Remove Bunch requirement

nsoranzo:
The `Bunch` package is unmaintained and not Python3-compatible, and is not used any more by CloudBridge.

Also, `config` needs to be any object accepted by `dict.update()`, fix the
documentation.


afgane:
Thanks for keeping up with this Nicola.

nuwang:
Excellent! One less dependency.

PR #186, opened 2019-02-14
[WIP] Subservices

almahmoud:


cloudbridge/cloud/base/resources.py
```diff
@@ -965,3 +884,7 @@ def wait_till_ready(self, timeout=None, interval=None):
             terminal_states=[GatewayState.ERROR, GatewayState.UNKNOWN],
             timeout=timeout,
             interval=interval)
+
+    def delete(self):
+        return self._provider.networking._gateways.delete(self.network_id,
```
nuwang: This and other places needs a ` # pylint:disable=protected-access` comment to indicate that we are doing this on purpose.

cloudbridge/cloud/base/services.py
```diff
@@ -62,6 +62,18 @@ class BaseSecurityService(SecurityService, BaseCloudService):
     def __init__(self, provider):
         super(BaseSecurityService, self).__init__(provider)
 
+    @property
+    def key_pairs(self):
+        return self._key_pairs
```
nuwang: this probably shouldn't be in the base class because the base class doesn't have a property called _key_pairs. Subclasses should probably redefine this despite being redundant because someone implementing a new provider may not notice that this is actually unimplemented.

cloudbridge/cloud/base/services.py
```diff
@@ -243,6 +259,26 @@ class BaseNetworkingService(NetworkingService, BaseCloudService):
     def __init__(self, provider):
         super(BaseNetworkingService, self).__init__(provider)
 
+    @property
+    def networks(self):
+        return self._network_service
```
nuwang: Same issue as above here.

cloudbridge/cloud/interfaces/services.py
```diff
@@ -965,7 +965,7 @@ def create(self, name, location=None):
         pass
 
 
-class BucketObjectService(PageableObjectMixin, CloudService):
+class BucketObjectService(CloudService):
```
nuwang: Shouldn't the bucket object service remain a pageable service?

cloudbridge/cloud/base/services.py
```diff
@@ -243,6 +259,26 @@ class BaseNetworkingService(NetworkingService, BaseCloudService):
...
+    def subnets(self):
+        return self._subnet_service
+
+    @property
+    def routers(self):
+        return self._router_service
+
+    @property
+    def _gateways(self):
+        return self._gateway_service
```
nuwang: these property should also be defined in the cloudbridge interface?

cloudbridge/cloud/providers/aws/services.py
```diff
@@ -1088,3 +1159,115 @@ def delete(self, router):
...
+    def get(self, gateway, fip_id):
+        log.debug("Getting AWS Floating IP Service with the id: %s", fip_id)
+        return self.svc.get(fip_id)
+
+    @dispatch(event="provider.networking.floating_ips.list",
+              priority=BaseFloatingIPService.STANDARD_EVENT_PRIORITY)
+    def list(self, gateway, limit=None, marker=None):
+        log.debug("Listing all floating IPs under gateway %s", gateway)
+        limit = limit or self.provider.config.default_result_limit
+        collection = self.boto_collection.filter()
```
nuwang: I'm not sure why we aren't just calling self.svc.list(). The only difference I can see from the svc implementation is in this line but that shouldn't have an effect?

cloudbridge/cloud/providers/openstack/services.py
```diff
@@ -1083,3 +1130,99 @@ def create(self, label, network):
...
+                         if n.external)
+        for net in external_nets:
+            if self._check_fip_connectivity(network, net):
+                return OpenStackInternetGateway(self._provider, net)
+        return None
+
+    @dispatch(event="provider.networking.gateways.delete",
+              priority=BaseGatewayService.STANDARD_EVENT_PRIORITY)
+    def delete(self, network, gateway):
+        pass
```
nuwang: Do we need a delete() method etc for gateways? The only method we seem to need is the get_or_create() method, because listing etc. doesn't make sense on most clouds. This would also cause tests to fail if we run the standard service tests on these, so I feel that this should not be a full fledged service.

cloudbridge/cloud/base/services.py
almahmoud: Ok I will move them back!

cloudbridge/cloud/interfaces/services.py
almahmoud: It is not iterable given that the list method needs a bucket parameter. The subservices are iterable.

cloudbridge/cloud/providers/aws/services.py
almahmoud: Because we need the gateway when wrapping the objects at 1238. The generic constructor with `cb_resource` doesn't work.

cloudbridge/cloud/providers/openstack/services.py
almahmoud: The delete method is needed for AWS where gateways are a real resource

cloudbridge/cloud/interfaces/services.py
nuwang: Ok makes sense...

PR #195, opened 2019-02-21
[WIP] Adding middleware support in factory

almahmoud:
This adds support for middleware in factory, which then gets inherited by any provider object created by this factory.

Three options to add middleware from factory:
1) Direct middleware add which will keep the same object across calls from any provider created by this factory after adding (in this example, the count will continue for all calls)
```

        class SomeDummyClass(object):
            count = start_count

            @intercept(event_pattern="*", priority=2499)
            def return_incremented(self, event_args, *args, **kwargs):
                self.count += 1
                return self.count

        some_obj = SomeDummyClass()
        factory.middleware.add(some_obj)
```

2) Middleware constructor add which will create the middleware with the class and arguments when making a new middleware manager for the provider object (in this example, count will only keep track of calls within each provider):
```
        class SomeDummyClassWithArgs(object):
            def __init__(self, start, increment):
                self.count = start
                self.increment = increment

            @intercept(event_pattern="*", priority=2499)
            def return_incremented(self, event_args, *args, **kwargs):
                self.count += self.increment
                return self.count

        factory = CloudProviderFactory()
        factory.middleware.add_constructor(SomeDummyClassWithArgs,
                                           start_count, increment)
```
and 3) adding events directly which will recreate a handler with the same callback function (bound methods will remain bound to their object)
```
        def return_hello(event_args, *args, **kwargs):
            return "hello"

        factory = CloudProviderFactory()
        factory.middleware.events.intercept("*", 2490, return_hello)
```

For all these scenarios, provider objects will inherit the events from the factory:
```
aws = factory.create_provider(ProviderList.AWS, {})
az = factory.create_provider(ProviderList.AZURE, {})
# Both providers will inherit all middleware/handlers added to the factory
```


cloudbridge/cloud/base/provider.py
```diff
@@ -83,11 +83,11 @@ def debug_mode(self):
 
 
 class BaseCloudProvider(CloudProvider):
-    def __init__(self, config):
+    def __init__(self, config, middleware_manager=None):
```
nuwang: This is great, better than sending in middleware as a list.

cloudbridge/cloud/factory.py
```diff
@@ -12,6 +16,54 @@
...
 
 
+# Todo: Move to pyeventsystem if we're keeping this logic
+class ParentMiddlewareManager(SimpleMiddlewareManager):
+
+    def __init__(self, event_manager=None):
+        super(ParentMiddlewareManager, self).__init__(event_manager)
+        self.middleware_constructors = []
+
+    def add_constructor(self, middleware_class, *args):
```
nuwang: This could possibly be renamed to add_middleware_class, accepting the class in question, and both the constructor args and kwargs.

cloudbridge/cloud/factory.py
```diff
@@ -12,6 +16,54 @@
...
+        for middleware in self.middleware_list:
+            new_manager.add(middleware)
+        for constructor, args in self.middleware_constructors:
+            m = constructor(*args)
+            new_manager.add(m)
+        for handler in self.get_subscribed_handlers():
+            new_handler = handler.__class__(handler.event_pattern,
+                                            handler.priority,
+                                            handler.callback)
+            new_manager.events.subscribe(new_handler)
```
nuwang: So this means that events added to the factory later won't be subscribed to previously created providers? This is ok, but I think we need to clarify how exactly factory event subscriptions are meant to work and what the intended use cases are, so we know for sure that this is the correct behaviour.

cloudbridge/cloud/factory.py
```diff
@@ -12,6 +16,54 @@
...
+            m = middleware
+        else:
+            m = AutoDiscoveredMiddleware(middleware)
+        self.middleware_list.append(m)
+        return m
+
+    def get_subscribed_handlers(self):
+        handlers = []
+        # Todo: Expose this better in pyeventsystem library
+        event_dict = self.events._SimpleEventDispatcher__events
```
nuwang: Not sure why we can't just access self.events?

cloudbridge/cloud/factory.py
```diff
@@ -27,9 +79,14 @@ class CloudProviderFactory(object):
     """
 
     def __init__(self):
+        self._middleware = ParentMiddlewareManager()
```
nuwang: I think we need a different name because ParentMiddlewareManager doesn't seem to explain what it's doing. Also, there is the issue of namespacing. And finally, I guess there's the issue of being able to intercept behaviour of a particular cloud (say openstack), and change behaviour for all classes of that cloud. One way would be to prefix the cloudname to the event. Another way would be is to check the instancetype of the sender: `if isinstance(OpenStackProvider, event_args['sender'].provider)` or some such thing. I think namespacing would be good as long as isolation is maintained.

cloudbridge/cloud/factory.py
```diff
@@ -12,6 +16,54 @@
...
+            new_handler = handler.__class__(handler.event_pattern,
+                                            handler.priority,
+                                            handler.callback)
+            new_manager.events.subscribe(new_handler)
+        return new_manager
+
+    # Removing install step from add. Since this manager is meant to create
+    # other managers rather than run. This will also simplify separating
+    # handlers added through middleware from those subscribed directly
+    def add(self, middleware):
```
nuwang: I don't think we should provide dual semantics here  - one for shared middleware objects, and another for independent middleware, just to keep it simple. I'm currently leaning in favour of the factory namespacing child event dispatchers, and forwarding them as necessary, so that we still maintain event isolation within a given provider. I think we should avoid a scenario where providers are unwittingly subscribing to events outside of the provider.

cloudbridge/cloud/factory.py
almahmoud: `self.events` is the `SimpleEventDispatcher`, which currently doesn't have a property exposing the events dict. We definitely should add the property, but I was avoiding changing the other library for the time being, before we discuss this further.

cloudbridge/cloud/factory.py
almahmoud: That was the original question that I had about this: whether we should make the provider object independent after it was created, or always treat it as a "child" of the factory. If the latter, it's definitely doable, but I would then suggest making the factory accessible through the provider as well.

PR #204, opened 2019-05-31
Enable passing GCP `project_id` in config.

VJalili:
The GCP backend expects the `project_id` in either credentials file/dict or from an environment variable. This PR adds a third option of passing the `project_id` via the `config` dictionary

### DIFF
--- a/cloudbridge/providers/gcp/provider.py
+++ b/cloudbridge/providers/gcp/provider.py
@@ -230,6 +230,8 @@ class GCPCloudProvider(BaseCloudProvider):
 
         if self.credentials_dict and 'project_id' in self.credentials_dict:
             self.project_name = self.credentials_dict['project_id']
+        elif self._get_config_value("project_id"):
+            self.project_name = self._get_config_value("project_id")
         else:
             self.project_name = os.environ.get('GCP_PROJECT_NAME')
 
--- a/docs/topics/setup.rst
+++ b/docs/topics/setup.rst
@@ -56,9 +56,9 @@ will override environment values.
     }
     config = {'gcp_service_creds_dict': gcp_creds}
     # A third alternative is to use a GCP credentials object provided by the GCP python
-    # sdk. This is for advanced usage scenarios.
-    # e.g. credentials = AccessTokenCredentials(access_token, "MyAgent/1.0", None)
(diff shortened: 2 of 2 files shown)


cloudbridge/providers/gcp/provider.py
```diff
@@ -230,6 +230,8 @@ def __init__(self, config):
 
         if self.credentials_dict and 'project_id' in self.credentials_dict:
             self.project_name = self.credentials_dict['project_id']
+        elif self._get_config_value("project_id"):
```
nuwang: It seems that project_id is already being read in from the credentials_dict, so it looks like we can just reuse that.
```
 config = {
    'gcp_credentials_obj': credentials,
    'gcp_credentials_dict': { 'project_id': '<project_id>'}
  }
```
This will work because credentials_obj always takes precedence over credentials_dict, so passing in both shouldn't cause problems.

cloudbridge/providers/gcp/provider.py
```diff
@@ -230,6 +230,8 @@ def __init__(self, config):
...
+            self.project_name = self._get_config_value("project_id")
```
nuwang: We can also leave this in, and perhaps simplify it as follows:
```
   self.project_name = self._get_config_value("project_id", os.getenv('GCP_PROJECT_NAME'))
```
If we do this, the only issue is that there will be 3 sources for project_id to come in from: the credentials_dict, a direct config value or the os environment, in that order. Any opinions @afgane @almahmoud 

Also, there's the question of why the ENV var and variable is called PROJECT_NAME, but the gcp nomenclature is project_id.

cloudbridge/providers/gcp/provider.py
VJalili: Correct; though a flat dictionary may read better than a nested one.

cloudbridge/providers/gcp/provider.py
VJalili: I guess if we follow your previous comment, there will be only two sources for `project_id`, and that makes sense where one is directly passed, and one read from an environment variable.

VJalili:
I am closing this since it seems we can leverage existing methods to specify project_id when a credentials object is used.

PR #208, opened 2019-06-20
AWS VM Types from auto-updated GitHub repo

almahmoud:
https://github.com/CloudVE/aws-instance-types seems to be updating properly with no obvious recurring errors, so I switched to using that as a source for vm_types, which makes them now filtered by availability zone.

### DIFF
--- a/cloudbridge/providers/aws/services.py
+++ b/cloudbridge/providers/aws/services.py
@@ -826,21 +826,23 @@ class AWSVMTypeService(BaseVMTypeService):
     @cachetools.cached(cachetools.TTLCache(maxsize=1, ttl=24*3600))
     def instance_data(self):
         """
-        Fetch info about the available instances.
+        Fetch info about the available VM types.
 
         To update this information, update the file pointed to by the
-        ``provider.AWS_INSTANCE_DATA_DEFAULT_URL`` above. The content for this
-        file should be obtained from this repo:
+        ``provider.AWS_INSTANCE_DATA_DEFAULT_URL``. The content for this
+        file can be obtained from this repo:
         https://github.com/powdahound/ec2instances.info (in particular, this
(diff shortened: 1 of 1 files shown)


cloudbridge/providers/aws/services.py
```diff
@@ -826,24 +826,37 @@ def __init__(self, provider):
...
-            self.provider.AWS_INSTANCE_DATA_DEFAULT_URL))
-        # Some instances are only available in certain regions. Use pricing
-        # info to determine and filter out instance types that are not
-        # available in the current region
-        vm_types_list = r.json()
-        return [vm_type for vm_type in vm_types_list
-                if vm_type.get('pricing', {}).get(self.provider.region_name)]
+        url = "https://raw.githubusercontent.com/CloudVE/aws-instance-types" \
+              "/master/vmtypes/{}.json".format(self.provider.zone_name)
+        r = requests.get(url)
```
nuwang: @almahmoud Should we consider pushing this to cloudfront for better performance perhaps?

cloudbridge/providers/aws/services.py
```diff
@@ -826,24 +826,37 @@ def __init__(self, provider):
...
+        vm_types_list = list(r.json())
+        if vm_types_list:
+            return vm_types_list
+        else:
```
nuwang: Why do we need this fallback to the previous method?

cloudbridge/providers/aws/services.py
almahmoud: Added push to S3 and now getting from S3 first, backup is github

cloudbridge/providers/aws/services.py
almahmoud: Changed fallback to github, first try from S3

PR #221, opened 2019-09-09
SubnetServices does not need network for get

almahmoud:


cloudbridge/base/subservices.py
```diff
@@ -142,7 +142,7 @@ def _provider(self):
         return self.__provider
 
     def get(self, subnet_id):
-        sn = self._provider.networking.subnets.get(self.network, subnet_id)
+        sn = self._provider.networking.subnets.get(subnet_id)
```
nuwang: Hmm... how was this not caught by our tests?

nuwang: Also, do we need to update the tests so this is caught?

nuwang: @almahmoud Should this PR be merged or should it wait for the test?

almahmoud: I'd say merge for now and create a separate issue for the test since this is a small bug fix and idk when I'll have time to do the test... I was planning on shifting to CloudBridge but that never happened cause other things kept coming up... :/

nuwang: Actually, now that I think about it, this may be somewhat inconsistent. In most other services which have a subservice, the first parameter is the containing object. E.g.: https://github.com/CloudVE/cloudbridge/blob/200e30940aac9001666fb73b505e4a77a6da621d/cloudbridge/interfaces/services.py#L1611
This doesn't affect us functionally, so we can fix this later if we need, but will open a separate issue to track.

PR #256, opened 2020-11-11
OS floating IP fix

almahmoud:


nuwang:
@almahmoud Looks like the original commit line with the fix has been dropped.

afgane (approved):

PR #257, opened 2020-11-16
Make instance creation on AWS more robust

nuwang:
Launching instances can result in the following error:
```
[2020-11-16 10:03:44,544: ERROR/ForkPoolWorker-2] Task cloudlaunch.tasks.create_appliance[07823fe4-169e-477e-8ca7-b3103e6dabbd] raised unexpected: Exception("Create appliance task failed: CloudBridgeBaseException: An error occurred (InvalidInstanceID.NotFound) when calling the CreateTags operation: The instance ID 'i-0c2fcd72eec4901f2' does not exist from exception type: <class 'botocore.exceptions.ClientError'>")
Traceback (most recent call last):
  File "/Users/Nuwan/work/cloudlaunch/venv/lib/python3.8/site-packages/cloudbridge/base/middleware.py", line 45, in wrap_exception
    return next_handler.invoke(event_args, *args, **kwargs)
  File "/Users/Nuwan/work/cloudlaunch/venv/lib/python3.8/site-packages/pyeventsystem/events.py", line 110, in invoke
    result = self.callback(*args, **kwargs)
  File "/Users/Nuwan/work/cloudlaunch/venv/lib/python3.8/site-packages/cloudbridge/providers/aws/services.py", line 785, in create
    inst[0].label = label
  File "/Users/Nuwan/work/cloudlaunch/venv/lib/python3.8/site-packages/cloudbridge/providers/aws/resources.py", line 259, in label
    self._ec2_instance.create_tags(Tags=[{'Key': 'Name',
  File "/Users/Nuwan/work/cloudlaunch/venv/lib/python3.8/site-packages/boto3/resources/factory.py", line 520, in do_action
    response = action(self, *args, **kwargs)
  File "/Users/Nuwan/work/cloudlaunch/venv/lib/python3.8/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(*args, **params)
  File "/Users/Nuwan/work/cloudlaunch/venv/lib/python3.8/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/Nuwan/work/cloudlaunch/venv/lib/python3.8/site-packages/botocore/client.py", line 676, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (InvalidInstanceID.NotFound) when calling the CreateTags operation: The instance ID 'i-0c2fcd72eec4901f2' does not exist
```

This happens in the following cloudbridge code, just after we create an instance:
```
            # Wait until the resource exists
            # pylint:disable=protected-access
            inst[0]._wait_till_exists()
            # Tag the instance w/ the name
            inst[0].label = label
```

It appears that boto code for waiting till an instance is ready does not work as expected. This seems to be an issue on the AWS API side - it looks like their API server replicas have a delay in syncing and as a result, there's no "read-after-write" consistency.

There's a large number of similar concurrency issues that have now sprung up. We may also need to get in touch with AWS support about this.

### DIFF
--- a/cloudbridge/providers/aws/services.py
+++ b/cloudbridge/providers/aws/services.py
@@ -782,7 +782,16 @@ class AWSInstanceService(BaseInstanceService):
             # pylint:disable=protected-access
             inst[0]._wait_till_exists()
             # Tag the instance w/ the name
-            inst[0].label = label
+            try:
+                inst[0].label = label
+            except Exception:
+                # It's possible for the label setter to fail, because EC2
+                # endpoints have a delay in syncing, and the instance may not
+                # yet be visible if a different endpoint is hit. To compensate,
+                # we retry in both the label setter, and raise here again just
+                # in case the label setter fails for any other reason.
+                inst[0].delete()
+                raise
             return inst[0]
         raise ValueError(
             'Expected a single object response, got a list: %s' % inst)
--- a/cloudbridge/providers/aws/services.py
+++ b/cloudbridge/providers/aws/services.py
@@ -10,6 +10,8 @@ import cachetools
 
 import requests
 
+import tenacity
+
 import cloudbridge.base.helpers as cb_helpers
 from cloudbridge.base.middleware import dispatch
 from cloudbridge.base.resources import ClientPagedResultList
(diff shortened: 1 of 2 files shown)


cloudbridge/providers/aws/services.py
```diff
@@ -782,7 +782,16 @@ def create(self, label, image, vm_type, subnet,
             # pylint:disable=protected-access
             inst[0]._wait_till_exists()
             # Tag the instance w/ the name
-            inst[0].label = label
+            try:
+                inst[0].label = label
```
afgane, line 786: The tests keep keep failing [here](https://travis-ci.com/github/CloudVE/cloudbridge/jobs/441076339#L1071). Should this be wrapped into a method with retry as well? Same for the security group label and description (https://github.com/CloudVE/cloudbridge/blob/144910ff7787d848a5cbcf9f22494a1984cb36f4/cloudbridge/providers/aws/services.py#L194)?

nuwang: I should have used `inst[0].delete()` instead of `inst.delete()`. Should be fixed now.

afgane: That fixed the instance labeling failures but setting the security group labels is facing the same issue now so should it be wrapped? 
More generally, as you keep saying, it really seems the `wait_till_ready` is not functioning as is should from the AWS side. This [failure to create an image](https://travis-ci.com/github/CloudVE/cloudbridge/jobs/441127560#L993) should not have happened tracing the calls... I guess more retries will need to be added for the time being.

nuwang: Have added a retry to every label setter. I think we need to think about a more comprehensive plan here. Maybe every operation should atomically succeed or fail. If it fails, it should presumably undo what it did, such as deleting a volume it just created.

afgane (approved):

PR #272, opened 2021-07-13
Fixing wrong logging object in AWS module files

MosheFriedland:
Before this change, it was not able to mute or change the logging levels of these files.
Now, they will belong to the cloudbridge module, so users will be able to do something like:
`logging.getLogger('cloudbridge').setLevel(logging.INFO)` for their own use :)


almahmoud (approved):
LGTM

nuwang:
@MosheFriedland Thanks for making these fixes!

PR #274, opened 2021-09-03
Update Azure dependencies to latest versions

### LINKED ISSUE #266: Upgrade Azure libraries
The provider libraries in the `setup.py` file are quite old versions.  Some have annoying issues that have since been fixed.  E.g.

https://github.com/Azure/azure-cli/pull/12839

Would it be possible to upgrade these dependencies to more recent versions?

FabioRosado:
Hello everyone, thank you for creating this great project. This PR aims to fix #266, I'm not sure if I wasn't too eager at updating all the dependencies to the latest version, also should these be pinned in a range between current up to the major one just in case?

I haven't manually tested this yet, but I am going to take it for a spin locally later today

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,8 +6,7 @@ import ast
 import os
 import re
 
-from setuptools import find_packages
-from setuptools import setup
+from setuptools import find_packages, setup
 
 # Cannot use "from cloudbridge import get_version" because that would try to
 # import the six package which may not be installed yet.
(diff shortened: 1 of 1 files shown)


nuwang:
@FabioRosado Thanks for making this PR! It's great that you're taking a stab at this. I do recall that the last time we tried to update the libraries, there were some changes to the azure interfaces, so the cloudbridge code may need some changes to match.

FabioRosado:
haha yeah I expected that somethings wouldn't work with the new versions, I'm going to have a look at these tests closer to see if I can figure out whats broken and I'll add new commits to this PR 😄

FabioRosado:
Started looking at the first error in the tests -  that `BlobPermissions` isn't found. Azure removed this and some methods like the `generate_blob_shared_access_signature` method.

I'm looking at the azure docs to see how to change this, as you mentioned I suspect that this is just the tip of the iceberg and perhaps a lot of the codebase for azure will need to be updated

nuwang:
@FabioRosado No worries. I've started a PR that upgrades other dependencies to their latest versions: https://github.com/CloudVE/cloudbridge/pull/275
Let me know if you're still interested in working on the Azure upgrade, it's not something we use at the moment, so any and all help would be appreciated.

nuwang:
@FabioRosado I've started some work on this. Have moved your commit to this PR: https://github.com/CloudVE/cloudbridge/pull/276 and have done some preliminary work for the update. The main thing outstanding at the moment is the blob service, which seems to have gone through some significant changes.

FabioRosado:
This is great thank you for working on this. I know it has taken me a while to get more work on this PR, would you like me to work on the blob service?

nuwang:
@FabioRosado That would be great. It looks like something is off with resourceGroups in general: https://github.com/CloudVE/cloudbridge/runs/3640265245?check_suite_focus=true so if you could take a look, that'd be great. I don't have proper access to an azure project, which I'll try to sort out on Monday, hence it's a bit difficult to debug things.

I'll close this PR in favour of #276. Please feel free to make additional PRs against that branch (update_azure_09_21), and we can run the integration tests.

FabioRosado:
Sure thing! My holidays are almost at the end, but I do have access to an Azure account that we use for dev at work. I'll give that a try and see if I can get more information about these failures :smile:

PR #277, opened 2021-09-19
Update Azure libs - part 2

FabioRosado:
This is my first stab at updating Azure dependencies following your awesome work @nuwang I have a bunch of tests still failing which I need to look into and I would like to do some manual testing just in case as well since a lot changed in the Azure API.

I'll be adding some comments/questions but feel free to skip this until I do more work on this PR :smile: 

I've also added some redundant changes to make typing happy, not sure if you care about it and I'm happy to revert these changes :smile:

### DIFF
--- a/cloudbridge/providers/azure/azure_client.py
+++ b/cloudbridge/providers/azure/azure_client.py
@@ -21,7 +21,8 @@ from azure.mgmt.network import NetworkManagementClient
 from azure.mgmt.resource import ResourceManagementClient
 from azure.mgmt.resource.subscriptions import SubscriptionClient
 from azure.mgmt.storage import StorageManagementClient
-from azure.storage.blob import BlobSasPermissions, BlobServiceClient
+from azure.storage.blob import (BlobSasPermissions, BlobServiceClient,
+                                generate_blob_sas)
 
 from . import helpers as azure_helpers
 
@@ -281,6 +282,9 @@ class AzureClient(object):
                 self.public_key_storage_table_name)
         return self._table_service
 
+    def blob_client(self, container_name, blob_name):
+        return self.blob_service.get_blob_client(container=container_name, blob=blob_name)
+
     def get_resource_group(self, name):
         return self.resource_client.resource_groups.get(name)
 
@@ -403,7 +407,7 @@ class AzureClient(object):
     def delete_vm_firewall_rule(self, fw_rule_id, vm_firewall):
         url_params = azure_helpers.parse_url(VM_FIREWALL_RULE_RESOURCE_ID,
                                              fw_rule_id)
-        name = url_params.get(VM_FIREWALL_RULE_NAME)
+        name = url_params.get(VM_FIREWALL_RULE_NAME, "")
         return self.network_management_client.security_rules. \
             begin_delete(self.resource_group, vm_firewall, name).result()
 
@@ -416,14 +420,13 @@ class AzureClient(object):
     def create_container(self, container_name):
         try:
             return self.blob_service.create_container(container_name)
-        except AzureConflictHttpError as cloud_error:
-            if cloud_error.error_code == "ContainerAlreadyExists":
-                msg = "The given Bucket name '%s' already exists. Please " \
-                      "use the `get` or `find` method to get a reference to " \
-                      "an existing Bucket, or specify a new Bucket name to " \
-                      "create.\nNote that in Azure, Buckets are contained " \
-                      "in Storage Accounts." % container_name
-                raise DuplicateResourceException(msg)
+        except ResourceExistsError:
+            msg = "The given Bucket name '%s' already exists. Please " \
+                    "use the `get` or `find` method to get a reference to " \
+                    "an existing Bucket, or specify a new Bucket name to " \
+                    "create.\nNote that in Azure, Buckets are contained " \
+                    "in Storage Accounts." % container_name
+            raise DuplicateResourceException(msg)
 
     def get_container(self, container_name):
         return self.blob_service.get_container_client(container_name)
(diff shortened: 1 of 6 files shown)


cloudbridge/providers/azure/azure_client.py
```diff
@@ -1,28 +1,27 @@
 import datetime
 import logging
-from io import BytesIO
+
+import tenacity
+from cloudbridge.interfaces.exceptions import (DuplicateResourceException,
```
FabioRosado, line 5: This was my automatic formatter, should I change it back?

cloudbridge/providers/azure/azure_client.py
```diff
@@ -181,13 +180,13 @@ def __init__(self, config):
         log.debug("azure subscription : %s", self.subscription_id)
 
     @property
-    @tenacity.retry(stop=tenacity.stop_after_attempt(5), reraise=True)
+    @tenacity.retry(stop=tenacity.stop.stop_after_attempt(5), reraise=True)
```
FabioRosado, line 183: pylance was complaining and this seems to make it happy :shrug:  I'm happy to revert the change

cloudbridge/providers/azure/azure_client.py
```diff
@@ -310,8 +311,9 @@ def _get_or_create_storage_account(self):
             try:
                 self._storage_account = \
                     self.get_storage_account(self.storage_account)
-            except CloudError as cloud_error:
-                if cloud_error.error.error == "ResourceNotFound":
+            except HttpResponseError as cloud_error:
```
FabioRosado, line 314: I suspect we might need to change all of the `CloudError` to `HttpResponseError` since this is what Azure seems to raise now

cloudbridge/providers/azure/azure_client.py
```diff
@@ -367,47 +368,47 @@ def list_vm_firewall(self):
...
-            create_or_update(self.resource_group, name,
-                             parameters).result()
+            begin_create_or_update(self.resource_group, name,
+                                   parameters).result()
 
     def update_vm_firewall_tags(self, fw_id, tags):
         url_params = azure_helpers.parse_url(VM_FIREWALL_RESOURCE_ID,
                                              fw_id)
-        name = url_params.get(VM_FIREWALL_NAME)
+        name = url_params.get(VM_FIREWALL_NAME, "")
```
FabioRosado, line 377: This is one of the many "make typing happy" kind of change, let me know if you want me to revert this

cloudbridge/providers/azure/azure_client.py
```diff
@@ -417,66 +418,73 @@ def list_containers(self, prefix=None, limit=None, marker=None):
...
 
     def get_blob(self, container_name, blob_name):
-        return self.blob_service.get_blob_properties(container_name, blob_name)
+        blob_client = self.blob_client(container_name, blob_name)
+        return blob_client.get_blob_properties(container_name, blob_name)
 
-    def create_blob_from_text(self, container_name, blob_name, text):
-        self.blob_service.create_blob_from_text(container_name,
-                                                blob_name, text)
+    def create_blob_from_text(self, container_name, blob_name, text, length=None):
```
FabioRosado, line 449: Azure mentions that if you pass the `length` keyword argument, it can improve performance, so I added this here, what do you think?

cloudbridge/providers/azure/azure_client.py
```diff
@@ -417,66 +418,73 @@ def list_containers(self, prefix=None, limit=None, marker=None):
...
+        self.upload_blob(container_name, blob_name, text, length)
 
-    def create_blob_from_file(self, container_name, blob_name, file_path):
-        self.blob_service.create_blob_from_path(container_name,
-                                                blob_name, file_path)
+    def create_blob_from_file(self, container_name, blob_name, file_path, length=None):
+        with open(file_path, 'rb') as data:
```
FabioRosado, line 453: The new API doesn't seem to implement a `create_blob_from_path` but we can use `blob_upload` to upload data. 

I'm worried about the lack of checking, should we handle cases where the path isn't there? Or the assumption that the path exists is okay?

cloudbridge/providers/azure/azure_client.py
```diff
@@ -417,66 +418,73 @@ def list_containers(self, prefix=None, limit=None, marker=None):
...
+            self.upload_blob(container_name, blob_name, data, length)
 
-    def delete_blob(self, container_name, blob_name):
-        self.blob_service.delete_blob(container_name, blob_name)
+    def delete_blob(self, container_name, blob_name, delete_snapshots="include"):
+        blob_client = self.blob_client(container_name, blob_name)
+        blob_client.delete_blob(delete_snapshots)
 
     def get_blob_url(self, container_name, blob_name, expiry_time):
```
FabioRosado, line 460: This is one of those changes that might work, but I haven't tested this yet.

cloudbridge/providers/azure/azure_client.py
```diff
@@ -417,66 +418,73 @@ def list_containers(self, prefix=None, limit=None, marker=None):
...
+        return url
 
     def get_blob_content(self, container_name, blob_name):
-        out_stream = BytesIO()
-        self.blob_service.get_blob_to_stream(container_name,
-                                             blob_name, out_stream)
-        return out_stream
+        blob_client = self.blob_client(container_name, blob_name)
+        out_stream = blob_client.download_blob()
+        return out_stream.readall()
```
FabioRosado, line 477: This is a blocking call, should we returning the [StorageStreamDownloader](https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.storagestreamdownloader?view=azure-python) and let users choose how to download the blob instead?

cloudbridge/providers/azure/azure_client.py
nuwang: This if fine if flake8 is fine with it. Please try running `tox -e lint`.

cloudbridge/providers/azure/azure_client.py
nuwang: May have been a change in the latest version of tenacity? Fine if it's happy.

cloudbridge/providers/azure/azure_client.py
nuwang: Makes sense. Maybe we can directly catch `ResourceNotFoundError`?

cloudbridge/providers/azure/azure_client.py
nuwang: I think it's fine, thanks for fixing it.

cloudbridge/providers/azure/azure_client.py
nuwang: Can we make this len(text) and compute it? The cloudbridge interface doesn't have a length parameter, so it seems like an implementation specific thing.
```
self.upload_blob(container_name, blob_name, text, length=len(text))
```

cloudbridge/providers/azure/azure_client.py
nuwang: My intuition is that the assumption is ok, because if it's not, the error message will be reasonable. This gets autowrapped in a CloudBridgeBaseException anyway, so it can still be caught in a generic way, but the error message should be pretty easily comprehensible?

cloudbridge/providers/azure/azure_client.py
nuwang: Is the blob_contents method being called anymore? The iter_contents() method is the "non-blocking" version, and it's returning chunks now: https://github.com/CloudVE/cloudbridge/pull/277/files/eafa0811e41b07c8c82b566a9b889c4918f1966c#diff-6c6fa7ef69cdd88282c8567237b70cdb7501937a130df4ce156014b746649419R214

nuwang (commented):
@FabioRosado Thanks for making these changes, they are all steps in the right direction. Would you mind reopening this PR against branch `update_azure_09_21`. That way, we can keep merging your changes and running tests. External PRs don't get access to cloud credentials, but branches do, so we can then get the tests to run.

FabioRosado:
Thank you so much for the review Nuwan, I'll address your comments and do as you requested to opening this PR against that branch.

I'm going to pick this work on the weekend, hopefully, we can get some good progress done :smile:

PR #276, opened 2021-09-18
Update to latest azure libs

nuwang:
xref: https://github.com/CloudVE/cloudbridge/pull/274

PR #279, opened 2021-10-03
Update Azure dependencies - Part III  (Blob and container)

FabioRosado:
Hello @nuwang  I got a new dev account for Azure and I am testing things again. Thank you for merging the previous PR, it seems that the PR has less failing tests than when I run locally which is good. 

I'll add some comments to this PR and I will try to test a bunch of other stuff to see if we can get Azure updated and working again. Thank you for the patience with this!

Also happy hacktoberfest!

### DIFF
--- a/cloudbridge/providers/azure/azure_client.py
+++ b/cloudbridge/providers/azure/azure_client.py
@@ -408,10 +408,10 @@ class AzureClient(object):
             begin_delete(self.resource_group, vm_firewall, name).result()
 
     def list_containers(self, prefix=None, limit=None, marker=None):
-        results = self.blob_service.list_containers(prefix=prefix,
-                                                    num_results=limit,
+        results = self.blob_service.list_containers(name_starts_with=prefix,
+                                                    results_per_page=limit,
                                                     marker=marker)
-        return (results.items, results.next_marker)
+        return results
 
     def create_container(self, container_name):
         try:
--- a/cloudbridge/providers/azure/services.py
+++ b/cloudbridge/providers/azure/services.py
@@ -2,63 +2,40 @@ import base64
 import logging
 import uuid
 
+import cloudbridge.base.helpers as cb_helpers
+from cloudbridge.base.middleware import dispatch
+from cloudbridge.base.resources import (ClientPagedResultList,
+                                        ServerPagedResultList)
+from cloudbridge.base.services import (BaseBucketObjectService,
+                                       BaseBucketService, BaseComputeService,
+                                       BaseFloatingIPService,
+                                       BaseGatewayService, BaseImageService,
+                                       BaseInstanceService, BaseKeyPairService,
+                                       BaseNetworkingService,
+                                       BaseNetworkService, BaseRegionService,
+                                       BaseRouterService, BaseSecurityService,
+                                       BaseSnapshotService, BaseStorageService,
+                                       BaseSubnetService,
+                                       BaseVMFirewallRuleService,
+                                       BaseVMFirewallService,
+                                       BaseVMTypeService, BaseVolumeService)
+from cloudbridge.interfaces.exceptions import (DuplicateResourceException,
+                                               InvalidParamException,
+                                               InvalidValueException)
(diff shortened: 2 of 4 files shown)


cloudbridge/providers/azure/services.py
```diff
@@ -544,7 +521,7 @@ def get(self, bucket_id):
     def list(self, limit=None, marker=None):
         buckets = [AzureBucket(self.provider, bucket)
                    for bucket
-                   in self.provider.azure_client.list_containers()[0]]
```
FabioRosado, line 547: The API is now returning a [ItemPaged object](https://docs.microsoft.com/en-us/python/api/azure-core/azure.core.paging.itempaged?view=azure-python)

cloudbridge/providers/azure/azure_client.py
```diff
@@ -408,10 +408,10 @@ def delete_vm_firewall_rule(self, fw_rule_id, vm_firewall):
             begin_delete(self.resource_group, vm_firewall, name).result()
 
     def list_containers(self, prefix=None, limit=None, marker=None):
-        results = self.blob_service.list_containers(prefix=prefix,
-                                                    num_results=limit,
+        results = self.blob_service.list_containers(name_starts_with=prefix,
+                                                    results_per_page=limit,
                                                     marker=marker)
-        return (results.items, results.next_marker)
```
FabioRosado, line 414: Azure is returning an [ItemPaged object](https://docs.microsoft.com/en-us/python/api/azure-core/azure.core.paging.itempaged?view=azure-python) which is an iterable, we can return that and let `services.list` do its thing, seems to work okay when I tried. Let me know if you are happy with this.

Out of curiosity how would we pas the prefix to `list_containers`?  if this is called always from `provider.storage.buckets.list()` we can't pass any arguments to it right?

nuwang (approved):
Thanks for continuing to hack away at this. Changes look good. Happy hacktoberfest! Let me know if you want to merge these changes and trigger the integration tests.

FabioRosado:
> Thanks for continuing to hack away at this. Changes look good. Happy hacktoberfest! Let me know if you want to merge these changes and trigger the integration tests.

That's great! Thank you so much for pointing this to me, I'll add the `prefix` to this PR.
I think something funky is happening with uploading to a blob, the tests failed with a permissions error both on #276 and locally. Not sure if this is actually a permissions error, we changed the permissions in our dev accounts. But I find it suspicious that its failing in both places :smile:

FabioRosado:
Apologies for another ping @nuwang, I've figure out the issue with resources not being found in Azure. For example I created a disk called "test-volume", but when I did `provider.storage.volumes.get("test-volume")` Azure returned a Not Found error.

But when I pass the full name ("test-volume-4dc58f") it works. I believe this was a recent-ish change in Cloudbridge where we are adding a random(?) string to the name of the resources.

Is there a way to get this random string at the end of the resource somehow? If not we could try running `disks.list` and then iterate over them and return the disk with the id :thinking:

nuwang:
The reason for adding the random id is because unlike AWS and OpenStack, Azure and GCP do not generate unique resource IDs when you create objects, and instead use the name specified by the user as the resource identifier. This creates compatibility problems between providers because when you create a volume for example, you must give it a unique name, whereas in AWS and OpenStack, a unique name is never guaranteed. The design rationale for this is documented [here](http://cloudbridge.cloudve.org/en/latest/topics/design_decisions.html#resource-identification-naming-and-labeling).

The solution to this was to append a random uuid at the end, so that `vol = vol_service.create(label="test")` always returns a unique resource id. The id property in the returned volume contains the actual underlying volume id. So doing something like this should work:
```
  vol = vol_service.create(label="myvol")
  vol_service.delete(vol.id)
```
If you want to find a volume by user specified name, then the find method comes in handy:
```
matching_vols = vol_service.find(label="myvol")
for vol in matching_vols:
   vol.delete()
```

FabioRosado:
That's great thank you so much for the explanation. I will see if we can use `.find` if not I will see how we are implementing it and do the same for the `.get` methods

PR #282, opened 2021-10-12
Fix issues with upload blob, blob url and snapshots

FabioRosado:
I've seeing some weird things locally and going through each bit is taking a bit too long :disappointed: 

This new PR fixes issues with the blob - I also add to add the `Storage Blob Contributor` role in order to get the objects inside a storage bucket.

Getting the signed URL was also broken so I fixed it in this PR. Creating a snapshot from a volume wasn't working anymore because `create_or_update` expects you to pass a snapshot class. This PR adds the Disk object to the API, but there are some issues, for example, I didn't yet figure out how to get the new tag into the snapshot so our fetching will be broken :disappointed: 

A note on the `AzureBucketObject`, when we are listing the objects, Azure returns a BlobProperties, but when we are getting it returns the blobclient. The easiest way to handle this was to add `self._blob_properties` to the class and use that instead

### DIFF
--- a/cloudbridge/providers/azure/azure_client.py
+++ b/cloudbridge/providers/azure/azure_client.py
@@ -526,11 +526,12 @@ class AzureClient(object):
         return self.compute_client.snapshots.get(self.resource_group,
                                                  snapshot_name)
 
-    def create_snapshot(self, snapshot_name, params):
+    def create_snapshot(self, snapshot_name, volume, params):
         return self.compute_client.snapshots.begin_create_or_update(
             self.resource_group,
             snapshot_name,
-            params
+            volume,
+            params=params
         ).result()
 
     def delete_snapshot(self, snapshot_id):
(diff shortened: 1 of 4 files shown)


FabioRosado:
No worries happy to :smile: 

Today I spent some time looking into the Azure API but couldn't figure out how to replace the tag in a single command, so I am calling `update_snapshot_tags` when calling `create_snapshot` and confirmed tha the new snapshot will have the right tag. Still feels a bit hacky though :disappointed: 

One thing that I think might be worth mentioning. I've ran a bunch of tests manually (or my understanding of the crud ones) and manually they all work fine. But when running them with `tox` they keep erroring with the error 'resource not found'. I suspect the reason for this, is because Azure actually takes a long time to create things like network resources, firewalls, etc.

Do you think the issue could be timing and perhaps we need to wait for the resource to be created? :thinking:  Haven't dug much into the tests yet tbh.

On an unrelated note, would you be open to adding the hacktoberfest topic to the project? I'm okay if you prefer not, I still have nightmares from last year spam PRs haha

PR #294, opened 2021-11-29
Allow users to create signed urls with write permissions

### LINKED ISSUE #293: Should we allow users to specify read/write signed urls?
In GCP I had to create a custom method to generate a signed URL for a storage bucket because Cloudbridge is generating read only URLs. 

The aim of this issue, is to start a discussion if we should allow users to specify the type of permissions when generating the signed URL, we could add it as a parameter to be read (default) or write?

FabioRosado:
Fixes #293 

This is a draft PR to allow users to generate signed URLs with write permissions. I was a bit unsure on what to name the parameter, I was torn between:
- permissions
- is_write
- allow_write

Initially, I was implementing an ENUM so users could specify their permissions, would that be preferable? I decided against it after giving it some thought since perhaps accepting a bool would make it easier to work with the `generate_url` method, but I'm happy with either as long as we have some docs for it.

One thing that is missing from this PR is testing, I'm planning to reuse the `test_generate_url` test that we have and push the content with the URL then check that it's there.

So far I've tested this with GCP and seems to work fine, but I want to do a bit more manual testing as well

### DIFF
--- a/cloudbridge/providers/aws/resources.py
+++ b/cloudbridge/providers/aws/resources.py
@@ -871,7 +871,7 @@ class AWSBucketObject(BaseBucketObject):
     def delete(self):
         self._obj.delete()
 
-    def generate_url(self, expires_in, writable):
+    def generate_url(self, expires_in, writable=False):
         if writable:
             return self._provider.s3_conn.meta.client.generate_presigned_post(
                 self._obj.bucket_name, self.id, ExpiresIn=expires_in
--- a/cloudbridge/providers/aws/resources.py
+++ b/cloudbridge/providers/aws/resources.py
@@ -871,7 +871,11 @@ class AWSBucketObject(BaseBucketObject):
     def delete(self):
         self._obj.delete()
 
-    def generate_url(self, expires_in):
+    def generate_url(self, expires_in, write):
+        if write:
+            return self._provider.s3_conn.meta.client.create_presigned_post(
+                self._obj.bucket_name, self.id, expiration=expires_in
+            )
         return self._provider.s3_conn.meta.client.generate_presigned_url(
             'get_object',
             Params={'Bucket': self._obj.bucket_name, 'Key': self.id},
--- a/cloudbridge/providers/aws/resources.py
+++ b/cloudbridge/providers/aws/resources.py
@@ -871,10 +871,10 @@ class AWSBucketObject(BaseBucketObject):
     def delete(self):
         self._obj.delete()
 
-    def generate_url(self, expires_in, write):
-        if write:
-            return self._provider.s3_conn.meta.client.create_presigned_post(
-                self._obj.bucket_name, self.id, expiration=expires_in
+    def generate_url(self, expires_in, writable):
+        if writable:
+            return self._provider.s3_conn.meta.client.generate_presigned_post(
+                self._obj.bucket_name, self.id, ExpiresIn=expires_in
             )
         return self._provider.s3_conn.meta.client.generate_presigned_url(
             'get_object',
--- a/cloudbridge/providers/azure/resources.py
+++ b/cloudbridge/providers/azure/resources.py
@@ -258,7 +258,7 @@ class AzureBucketObject(BaseBucketObject):
         """
         self._blob_client.delete_blob()
 
-    def generate_url(self, expires_in, writable):
+    def generate_url(self, expires_in, writable=False):
         """
         Generate a URL to this object.
         """
--- a/cloudbridge/providers/azure/resources.py
+++ b/cloudbridge/providers/azure/resources.py
@@ -258,12 +258,12 @@ class AzureBucketObject(BaseBucketObject):
         """
         self._blob_client.delete_blob()
 
-    def generate_url(self, expires_in):
+    def generate_url(self, expires_in, write):
         """
         Generate a URL to this object.
         """
         return self._provider.azure_client.get_blob_url(
-            self._container, self.name, expires_in)
+            self._container, self.name, expires_in, write)
 
     def refresh(self):
         pass
--- a/cloudbridge/providers/azure/resources.py
+++ b/cloudbridge/providers/azure/resources.py
@@ -258,12 +258,12 @@ class AzureBucketObject(BaseBucketObject):
         """
         self._blob_client.delete_blob()
 
-    def generate_url(self, expires_in, write):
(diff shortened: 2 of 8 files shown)


nuwang:
I was going to suggest `writable` as an alternative, but this all looks great to me.

tests/test_object_store_service.py
```diff
@@ -193,6 +193,28 @@ def test_generate_url(self):
...
+                obj.upload(content)
+                target_stream = BytesIO()
+                obj.save_content(target_stream)
+
+                url = obj.generate_url(100, write=True)
+                if isinstance(self.provider, TestMockHelperMixin):
+                    raise self.skipTest(
+                        "Skipping rest of test - mock providers can't"
+                        " access generated url")
+                self.assertEqual(requests.post(url).content, content)
```
nuwang, line 216: I think this test should upload some new content to the signed url, then retrieve that new content separately using `test_bucket.objects.get()` and then compare?

FabioRosado:
I like writable! I’ll change it thank you!

FabioRosado:
Finally got time to keep working on this and finished off the implementation of #293, I've manually tested this in GCP and AWS and works as expected.

I don't have an OpenStack account so I can't test it fully and need to get new creds for Azure to test it with Azure but from the three Azure might work fine since its just a permissions thing that we need to pass when generating the URL :smile: )

cloudbridge/providers/aws/resources.py
```diff
@@ -871,7 +871,11 @@ def upload_from_file(self, path):
     def delete(self):
         self._obj.delete()
 
-    def generate_url(self, expires_in):
+    def generate_url(self, expires_in, writable):
```
nuwang, line 874:
```suggestion
    def generate_url(self, expires_in, writable=False):
```

cloudbridge/providers/azure/resources.py
```diff
@@ -258,12 +258,12 @@ def delete(self):
         """
         self._blob_client.delete_blob()
 
-    def generate_url(self, expires_in):
+    def generate_url(self, expires_in, writable):
```
nuwang, line 261:
```suggestion
    def generate_url(self, expires_in, writable=False):
```

cloudbridge/providers/gcp/resources.py
```diff
@@ -1975,14 +1975,21 @@ def delete(self):
              .delete(bucket=self._obj['bucket'], object=self.name)
              .execute())
 
-    def generate_url(self, expires_in):
+    def generate_url(self, expires_in, writable):
```
nuwang, line 1978:
```suggestion
    def generate_url(self, expires_in, writable=False):
```

cloudbridge/providers/openstack/resources.py
```diff
@@ -1316,7 +1316,8 @@ def delete(self):
                 result = result and del_res['success']
         return result
 
-    def generate_url(self, expires_in):
+    def generate_url(self, expires_in, writable):
```
nuwang, line 1319:
```suggestion
    def generate_url(self, expires_in, writable=False):
```

tests/test_object_store_service.py
```diff
@@ -193,6 +195,34 @@ def test_generate_url(self):
...
+                if isinstance(self.provider, TestMockHelperMixin):
+                    raise self.skipTest(
+                        "Skipping rest of test - mock providers can't"
+                        " access generated url")
+                elif isinstance(self.provider, GCPCloudProvider):
+                    requests.put(url, data=content)
+                elif isinstance(self.provider, AWSCloudProvider):
+                    requests.put(url['url'], data=url['fields'], files={"file": (obj_name, content)})
+                else:
+                    requests.post(url, data=content)
```
nuwang, line 220:
```suggestion
                else:
                    requests.put(url, data=content)
```

tests/test_object_store_service.py
```diff
@@ -12,6 +12,8 @@
 from cloudbridge.interfaces.provider import TestMockHelperMixin
 from cloudbridge.interfaces.resources import Bucket
 from cloudbridge.interfaces.resources import BucketObject
+from cloudbridge.providers.aws.provider import AWSCloudProvider
+from cloudbridge.providers.gcp import GCPCloudProvider
```
nuwang, line 16:
```suggestion
```

tests/test_object_store_service.py
nuwang: @FabioRosado I was just trying to follow why the AWS provider's request.put has more parameters? The docs suggest that `requests.put(url, data=content)` should also work?

nuwang (dismissed):

nuwang:
Thanks for making these changes so quickly @FabioRosado. I'll merge this in and prep for release.

tests/test_object_store_service.py
FabioRosado: That's because AWS doesn't return a single URL like GCP. It returns a dictionary containing: `url - endpoint` and `fields - all the needed data to get things uploaded`.

This is the [docs that I was following](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-presigned-urls.html?highlight=generate_presigned_url#generating-a-presigned-url-to-upload-a-file) were you looking at the same ones?

nuwang: @FabioRosado I was actually looking at this: https://docs.aws.amazon.com/code-samples/latest/catalog/python-s3-s3_basics-presigned_url.py.html
I guess the core problem here is that we'll need to have a consistent return value across all providers, so that cloud agnosticism is maintained. If the above code sample doesn't work, maybe the solution is to always return a dict instead? On everything other than AWS, `fields`, and `file` would be empty?

FabioRosado: I agree with you that its not good that we can't have a consistent return value across providers. The annoying thing with AWS is that they want these fields in the body of the request while the others give you a single url that you can use.

We could change it as you suggest, but wouldn't be confusing that a get method returns a string and a post one returns a dict?

nuwang: @FabioRosado It looks like using `generate_presigned_url` would make this consistent?
https://stackoverflow.com/questions/65198959/aws-s3-generate-presigned-url-vs-generate-presigned-post-for-uploading-files

nuwang: These changes make the tests pass as expected: https://github.com/CloudVE/cloudbridge/pull/295

PR #296, opened 2022-02-01
Add automatic docker image building

almahmoud:
image
Creates docker images with tags:
```
3.9-aws
3.9-azure
3.9-full
3.9-gcp
3.9-openstack
dev-3.9-aws
dev-3.9-azure
dev-3.9-full
dev-3.9-gcp
dev-3.9-openstack
```

### DIFF
--- a/scripts/bucket_upload.py
+++ b/scripts/bucket_upload.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python3
+
+import sys
+import argparse
+
+from cloudbridge.factory import CloudProviderFactory
+from cloudbridge.factory import ProviderList
+from cloudbridge.interfaces.exceptions import DuplicateResourceException
+
+def main(arguments):
+    parser = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument('filepath', help="File to upload", type=str)
+    parser.add_argument('destination', help="Destination in bucket", type=str)
+    parser.add_argument('-b', '--bucket', help="Name of destination bucket", type=str)
+    parser.add_argument('-p', '--provider',
+                        choices=[ProviderList.__dict__[p] for p in ProviderList.__dict__.keys() if all(x not in p for x in ['_', 'MOCK'])],
+                        help='Name of provider to use')
+    parser.add_argument('-r', '--retries', help="Maximum number of retries. Default = 5", type=int, default=5)
+    args = parser.parse_args(arguments)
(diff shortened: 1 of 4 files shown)


nuwang:
What's the expected use case for this? To be able to run a quick ad-hoc script? If so, I'd trim this down to just one image with all providers installed.

almahmoud:
My immediate usecase is making Github actions that push to SWIFT object store. I'm fine collapsing to have full across the board, I would highly suggest leaving dev vs slim though, `dev` is helpful for trying things out manually, but size of `slim` is ~8x smaller. More generally, if this doesn't seem desirable as something to maintain, also happy to leave this just in fork as well.

scripts/bucket_upload.py
```diff
@@ -0,0 +1,37 @@
...
+
+def main(arguments):
+    parser = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument('filepath', help="File to upload", type=str)
+    parser.add_argument('destination', help="Destination in bucket", type=str)
+    parser.add_argument('-b', '--bucket', help="Name of destination bucket", type=str)
+    parser.add_argument('-p', '--provider',
+                        choices=[ProviderList.__dict__[p] for p in ProviderList.__dict__.keys() if all(x not in p for x in ['_', 'MOCK'])],
```
nuwang, line 18: Perhaps it's time we make ProviderList an enum? https://stackoverflow.com/questions/29503339/how-to-get-all-values-from-python-enum-class/29503414

I think this is just a legacy thing from python 2 days.

scripts/bucket_upload.py
```diff
@@ -0,0 +1,37 @@
...
+    try:
+        bucket = provider.storage.buckets.create(args.bucket)
+    except DuplicateResourceException:
+        bucket = provider.storage.buckets.get(args.bucket)
+    ob = bucket.objects.create(args.destination)
+    uploaded = False
+    count = 0
+    while not uploaded:
+        print(f'Trying for the {count}th time')
+        uploaded = ob.upload_from_file(args.filepath)
```
nuwang, line 32: This seems to have exposed another bug in the API. Not all implementations return a bool here, only openstack and azure do. I think the correct behaviour should be to raise an exception if the upload fails, instead of returning a bool value.

nuwang (commented):
I think we can merge this, but shall we then slim this down to just 'dev' and 'slim'? I think that fewer is better in this instance.

Also, we may be able to get rid of slim too if the docker image is cached on github?

PR #298, opened 2022-07-28
Add Azure stop VM instance method

rodrigonull:
Add a method to stop VM instances running in Azure.


nuwang:
@rodrigonull Thanks for this contribution. I just noticed that restart_vm also calls wait(), but it would be great if neither method blocks. Will probably need some interface refactoring in future, will file an issue for that.

nuwang:
Released in v3.1.0

PR #299, opened 2022-07-28
Add VM instance create time property for all providers

rodrigonull:
Add the ability to retrieve the VM instance creation time for all supported providers.

### DIFF
--- a/tests/test_compute_service.py
+++ b/tests/test_compute_service.py
@@ -168,6 +168,9 @@ class CloudComputeServiceTestCase(ProviderTestBase):
             self.assertEqual(len(find_zone), 1,
                              "Instance's placement zone could not be "
                              " found in zones list")
+            self.assertIsNotNone(
+                test_instance.create_time,
+                "Instance must have it's creation time")
 
     @helpers.skipIfNoService(['compute.instances', 'compute.images',
                               'compute.vm_types'])
--- a/tests/test_compute_service.py
+++ b/tests/test_compute_service.py
@@ -1,3 +1,4 @@
+import datetime
 import ipaddress
 
 import six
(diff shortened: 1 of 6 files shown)


tests/test_compute_service.py
```diff
@@ -168,6 +168,9 @@ def test_instance_properties(self):
             self.assertEqual(len(find_zone), 1,
                              "Instance's placement zone could not be "
                              " found in zones list")
+            self.assertIsNotNone(
+                test_instance.create_time,
+                "Instance must have it's creation time")
```
nuwang, line 173:
```suggestion
                "Instance must have its creation time")
```

nuwang: Can we also add a test here to make sure that the returned object is always a python datetime object?

nuwang (commented):
@rodrigonull Thanks for this contribution. Do all the methods return a python datetime object? Can we add a test for that?

rodrigonull:
> @rodrigonull Thanks for this contribution. Do all the methods return a python datetime object? Can we add a test for that?

Only AWS and Azure return a datetime object. GCP and OpenStack return an ISO 8601 string. I was wondering if we should convert the string to datetime, but that probably would require to import a third party library, like [dateutil](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse), because I think only in python 3.7 a [method](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat) was added that allows to convert a ISO 8601 string to datetime. What do you think?

nuwang:
I think only in python 3.7 a [method](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat) was added that allows to convert a ISO 8601 string to datetime. What do you think?

I think we could go for that, since support for python<3 has already been dropped, and most people are likely above 3.7. Alternatively, we could return an ISO 8601 string if that's easier.

nuwang (approved):
Thanks, this looks good!

nuwang:
Released in v3.1.0

Issue #306, opened 2022-08-28
Gateway detection problem in JS2 from router quota 1


almahmoud:
We are currently using a dummy router to check for FIP connectivity. This particularly comes up when trying to get a gateway from a network eg: `list(net.gateways)[0]`. This method won't work well in JS2 because of the new default quota of 1 on routers (https://docs.jetstream-cloud.org/general/policies/#router-quota-policies), making the list gateways method not work unless you have 0 routers.
https://github.com/CloudVE/cloudbridge/blob/d51933a28677e4f2cc2c077fed036c7851766145/cloudbridge/providers/openstack/services.py#L1236

@nuwang I am wondering if the last comment in the bug report https://bugs.launchpad.net/neutron/+bug/1743480 can be used instead? I.e. a python equivalent of the CLI `openstack network list --external --no-share`?

nuwang:
@almahmoud  It sure looks like we just remained with the original workaround and never got around to trying the equivalent of `openstack network list --external --no-share`. The workaround is truly horrendous, and if we can replace it, that would be great. Do you know what the equivalent API call would be? Perhaps we can trace the HTTP calls to find out?

almahmoud:
I believe this is all we need to return the external non-shared network as a gateway https://github.com/CloudVE/cloudbridge/pull/307

vineetjain27:
hey @almahmoud I want to work on this please accept my contribution to this issue.

almahmoud:
Hey @vineetjain27, thank you for your interest, but I’ve already resolved this issue as it was time sensitive for internal use. We’d welcome contributions for some of the other open issues though!

PR #307, opened 2022-09-06
Use external non-shared network for gateway

almahmoud:
Addresses https://github.com/CloudVE/cloudbridge/issues/306

### DIFF
--- a/cloudbridge/providers/openstack/services.py
+++ b/cloudbridge/providers/openstack/services.py
@@ -1248,7 +1248,7 @@ class OpenStackGatewayService(BaseGatewayService):
         log.debug("OpenStack listing of all current internet gateways")
         igl = [OpenStackInternetGateway(self._provider, n)
                for n in self._provider.networking.networks
-               if n.external and self._check_fip_connectivity(network, n)]
+               if n.external and not n.shared]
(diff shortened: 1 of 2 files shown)


cloudbridge/providers/openstack/services.py
```diff
@@ -1249,7 +1249,7 @@ def get_or_create(self, network):
         external_nets = (n for n in self._provider.networking.networks
                          if n.external)
         for net in external_nets:
-            if self._check_fip_connectivity(network, net):
```
nuwang, line 1252: We can remove the `_check_fip_connectivity` related functions entirely?

almahmoud: We could, I just thought having the function might not hurt if we ever want to check for connectivity for other purposes?

nuwang: Doesn't really do anything of significance I think: https://github.com/CloudVE/cloudbridge/blob/d51933a28677e4f2cc2c077fed036c7851766145/cloudbridge/providers/openstack/services.py#L1229
Let's remove and add back if necessary.

almahmoud: Done

nuwang (approved):

PR #313, opened 2023-09-04
Add new `ec2_retries_value` config for  `AWSCloudProvider`

MosheFriedland:
As a user with lots of parallel boto3 requests, I would like to be able and modify the ec2 retries value, in case of some AWS issues


nuwang (approved):
Thanks for this enhancement! Looks great!

PR #330, opened 2026-05-19
Use a trusted publisher when publishing to PyPI

ksuderman:
This PR implements trusted publishing for PyPI uploads to enhance security and reduce the risk of supply chain attacks. The changes replace the traditional API token-based authentication with OpenID Connect (OIDC) token-based authentication provided by GitHub Actions.

The deployment workflow has been updated to use the trusted publishing mechanism where GitHub Actions generates short-lived OIDC tokens that PyPI can verify and trust. This eliminates the need to store long-lived API tokens as repository secrets, reducing the attack surface and potential for credential compromise.

**PyPI Configuration Required:** After merging this PR, repository administrators must configure trusted publishing on both PyPI and Test PyPI before the next release:

**For Production PyPI:**
1. Visit https://pypi.org
2. Login with maintainer credentials
3. Navigate to the `cloudbridge` project
4. Click "Manage" and then click "Publishing"
5. Fill in the form:
   - **Owner**: `CloudVE`
   - **Repository name**: `cloudbridge`
   - **Workflow filename**: `deploy.yaml`
   - **Environment name**: (leave blank)
6. Click "Add"

**For Test PyPI:**
1. Visit https://test.pypi.org
2. Login with maintainer credentials
3. Repeat the same configuration as above

**Post-Configuration:**
- Test the setup by creating a new pre-release tag
- Monitor the workflow execution to ensure trusted publishing works
- Remove the old API token secrets (`PYPI_API_TOKEN` and `TEST_PYPI_API_TOKEN`) from repository secrets once confirmed working
- If issues arise, the workflow can be temporarily reverted and API tokens restored

The trusted publishing setup provides better security, audit trails, and eliminates the need for credential rotation while maintaining the same publishing functionality.

Closes #329

PR #332, opened 2026-05-20
CI hygiene: split deploy job + dedicated env, skip CI on docs-only changes

nuwang:
Two unrelated CI tweaks bundled together.

## 1. Tighten trusted-publishing blast radius (follow-up to #330). ping @nsoranzo
Addresses https://github.com/CloudVE/cloudbridge/pull/330#issuecomment-4491898064.

- Split `deploy.yaml` into `build_packages` (no special permissions, uploads the sdist/wheel as an artifact) and `publish` (downloads and uploads). Only `publish` requests `id-token: write`, so the OIDC token can no longer be minted during the build step.
- Pin the `publish` job to a dedicated `pypi` GitHub environment so the trusted-publisher binding (and any future protection rules / environment-scoped secrets) is isolated to that environment rather than the whole repo.

PyPI and Test PyPI trusted-publisher configuration already updated.

## 2. Skip integration CI on docs-only changes
Adds `paths-ignore` for `docs/**`, `**.md`, `**.rst`, and `LICENSE` on every trigger in `integration.yaml` and `integration-cloud.yaml`. Doc-only PRs and pushes no longer burn lint, mock-provider, and cloud-provider test minutes.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -6,13 +6,9 @@ on:
     tags:
       - '*'
 
-permissions:
-  contents: read
-  id-token: write  # Required for trusted publishing to PyPI
-
 jobs:
-  build-n-publish:
-    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+  build_packages:
+    name: Build Python 🐍 distributions 📦
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


ksuderman (approved):
LGTM

nsoranzo:
Thanks @nuwang !

